PR #13, opened 2026-03-13
Do not start tasks with uncompleted dependencies

BarreiroT:
Tasks with dependencies should wait for their dependencies to be completed before starting. This PR completes this and adds test for this case.


saoudrizwan:
ohh i completely misunderstood the PR - this is for the 'start all tasks' button in the backlog column header?

BarreiroT:
> ohh i completely misunderstood the PR - this is for the 'start all tasks' button in the backlog column header?

Hey Saoud, that is correct. This PR ensures that whenever you start all the tasks, only those with no pending dependencies start. You can still start them manually, or once they're done, but we won't start them alltogether.

PR #48, opened 2026-03-21
Normalize user input when rendering it [ENG-1658]

BarreiroT:


BarreiroT:
Yeah... but that would mean having to modify what we sent vs what we have in the chat history. I think his lives better on the render than elsewhere

BarreiroT:
Yep, I just opened the PR to run tests in CI. Should've done it as draft. my bad.

PR #49, opened 2026-03-23
Provide Feedback For Errors for In Progress Items

shey-cline:
- Shows error state for in progress items
- Wrap error text to prevent overflowing

**Before**
Screenshot 2026-03-23 at 9 49 25 AM
Screenshot 2026-03-23 at 9 50 02 AM

**After**

Screenshot 2026-03-23 at 9 48 55 AM

Screenshot 2026-03-23 at 9 50 14 AM

### DIFF
--- a/web-ui/src/components/detail-panels/cline-chat-message-item.tsx
+++ b/web-ui/src/components/detail-panels/cline-chat-message-item.tsx
@@ -159,7 +159,7 @@ export function ClineChatMessageItem({ message }: { message: ClineChatMessage })
 	}
 	const label = message.role === "status" ? "Status" : "System";
 	return (
-		<div className="max-w-[85%] rounded-md border border-border bg-surface-3/70 px-3 py-2 text-sm whitespace-pre-wrap text-text-secondary">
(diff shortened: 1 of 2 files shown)


web-ui/src/components/detail-panels/cline-chat-message-item.tsx
```diff
@@ -159,7 +159,7 @@ export function ClineChatMessageItem({ message }: { message: ClineChatMessage })
 	}
 	const label = message.role === "status" ? "Status" : "System";
 	return (
-		<div className="max-w-[85%] rounded-md border border-border bg-surface-3/70 px-3 py-2 text-sm whitespace-pre-wrap text-text-secondary">
+		<div className="max-w-[85%] rounded-md border border-border bg-surface-3/70 px-3 py-2 text-sm whitespace-pre-wrap break-all text-text-secondary">
```
BarreiroT, line 162: WDYT of using `break-normal`? Have we seen errors with no spaces anywhere?

shey-cline: `break-words` is probably more appropriate. `break-normal` won't help because the long JSON/URL strings have no natural break points and would overflow and `break-all` is overkill

shey-cline: though we use break-all everywhere else in the file 😕

PR #98, opened 2026-03-27
Force Claude hyperlinks in Kanban

michaellatman:
## Summary

Enable Claude's built-in hyperlink output in Kanban by setting `FORCE_HYPERLINK=1` for Claude task sessions.

## Why

Claude gates custom-label OSC 8 links behind terminal capability detection. Kanban identifies itself as `TERM_PROGRAM=kanban`, which Claude does not recognize for that feature, so markdown/custom-label links were not emitted inside Kanban even though xterm can render them.

Using Claude's supported override keeps the fix narrow and avoids spoofing a different terminal.

## Impact

Newly launched Claude sessions in Kanban should emit OSC 8 hyperlinks for markdown/custom-label links, which lets the Kanban terminal render them as clickable links.

## Validation

- `npm run typecheck`
- `npm run web:typecheck`
- `npm run build`
- `npm run check` (via pre-commit hook)

### DIFF
--- a/.clinerules/workflows/release.md
+++ b/.clinerules/workflows/release.md
@@ -31,7 +31,35 @@ Or if no tags exist:
 git log --oneline --no-merges
 ```
 
-### 3. Curate the changelog
+### 3. Check for Cline SDK version changes
+
+Compare the `@clinebot/core` version in `package.json` at the last tag vs HEAD:
+
+```bash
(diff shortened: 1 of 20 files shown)


saoudrizwan (approved):

saoudrizwan:
thank you so much! lgtm ❤️

PR #108, opened 2026-03-28
Feat/OpenAI compatible endpoints

jl-codes:


f0o:
@saoudrizwan worth noting that OpenAI Compatible APIs are not supported still, there are countless issue reports that it is broken in one or many ways. The latest build (kanban@0.1.70) uses now `openai-native` as slug but it is just as non-functional still.

Maybe worth backporting the PR if it works?

PR #122, opened 2026-03-28
Guard oversized workspace diffs in worker detail view

### LINKED ISSUE #121: Guard worker diff rendering when a worktree has too many changes
## Summary
Opening a worker detail view can crash the Kanban server when the worktree is extremely dirty.

## Reproduction
1. Create or reuse a task worktree with a very large number of changed files or a very large total diff payload.
2. Open the worker detail view.
3. Let the default `All Changes` diff load.

## Actual
The detail view requests the full working-copy diff, the server materializes the full `{ oldText, newText }` payload for every changed file, and the process can run out of memory or become unstable.

## Expected
Kanban should refuse to build/render oversized workspace diffs and should show a clear UI state instead of crashing.

## Notes
The current detail view polls working-copy diffs continuously, so the failure repeats once the panel is opened.

nehaturov-hue:
## Summary
- add a hard server-side guard for oversized workspace diff payloads
- return a truncated workspace-changes response instead of materializing unbounded diff text
- show a clear "too many changes to render" state in the worker detail UI

## Problem
Opening a worker detail view on a very dirty worktree could force Kanban to build and poll a massive working-copy diff. That could destabilize or crash the server instead of degrading gracefully.

## Changes
- extend the workspace-changes response contract with truncation metadata
- stop building full diff payloads when the change set exceeds file-count or text-size safety limits
- render a non-crashing fallback panel when the diff is intentionally truncated
- add UI coverage for the truncated-diff state

## Verification
- `npm --prefix /var/home/ut93/ai_test/public_repos/kanban run test -- web-ui/src/components/card-detail-view.test.tsx test/runtime/trpc/workspace-api.test.ts`
- `npm --prefix /var/home/ut93/ai_test/public_repos/kanban run web:test -- card-detail-view`
- `npm --prefix /var/home/ut93/ai_test/public_repos/kanban run build`

Fixes cline/kanban#121

### DIFF
--- a/src/core/api-contract.ts
+++ b/src/core/api-contract.ts
@@ -36,6 +36,9 @@ export const runtimeWorkspaceChangesResponseSchema = z.object({
 	repoRoot: z.string(),
 	generatedAt: z.number(),
 	files: z.array(runtimeWorkspaceFileChangeSchema),
+	totalFileCount: z.number().int().nonnegative().optional(),
+	truncated: z.boolean().optional(),
+	warning: z.string().nullable().optional(),
 });
 export type RuntimeWorkspaceChangesResponse = z.infer<typeof runtimeWorkspaceChangesResponseSchema>;
 
--- a/src/workspace/get-workspace-changes.ts
+++ b/src/workspace/get-workspace-changes.ts
@@ -9,6 +9,8 @@ import type {
 import { getGitStdout } from "./git-utils";
 
 const WORKSPACE_CHANGES_CACHE_MAX_ENTRIES = 128;
+const MAX_WORKSPACE_CHANGE_FILES = 250;
+const MAX_WORKSPACE_CHANGE_TEXT_BYTES = 5_000_000;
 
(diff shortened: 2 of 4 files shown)


koriyoshi2041 (commented):
The file-count guard runs before content is loaded, but the text-byte guard still runs only *after* `await build()`. Each `buildFileChange*` call has already read the complete old/new file contents at that point (`readWorkingTreeFile`, `readHeadFile`, or `readFileAtRef`).

That means a worktree with only one very large tracked or untracked text file can still materialize arbitrarily large strings before the 5 MB limit is checked, leaving the original memory-pressure path open for low file counts.

Could the implementation preflight/cap content before those reads (for example, filesystem size for working-tree files and `git cat-file -s` or a bounded read for ref content), and add a server-side regression covering one oversized file? The PR description mentions `test/runtime/trpc/workspace-api.test.ts`, but the current diff only adds the UI truncated-state test, so this boundary does not appear to be exercised yet.

PR #147, opened 2026-03-30
fix: prevent non-ASCII file paths from rendering as octal escapes in diff view

### LINKED ISSUE #146: Non-ASCII (e.g. Japanese) file/directory names render as octal escape sequences in diff view
### What happened?

When a repository contains files or directories with non-ASCII characters (e.g., Japanese), the diff view in the Kanban web UI displays paths as octal escape sequences instead of readable text.

For example, a directory named `提出書類` is displayed as `\345\217\220\345\207\272\346\233\270\351\241\236`.

Image

**Root cause:** Git defaults to `core.quotepath=true`, which encodes non-ASCII bytes as octal escapes in path output. Several git commands in the codebase are invoked without `-z` (null-byte delimiter) or `-c core.quotepath=false`, so the escaped paths are passed through to the UI verbatim.

**Affected code paths:**

1. `src/workspace/get-workspace-changes.ts` (line 373) — `git diff --name-status HEAD --` is called without `-z`. Non-ASCII paths in the tab-delimited output are wrapped in double quotes with octal escapes (e.g., `"path/\345\217\220..."`) and `parseTrackedChanges()` does not strip quotes or decode escapes.

2. `src/workspace/get-workspace-changes.ts` (line 374) — `git ls-files --others --exclude-standard` is called without `-z`. Paths with non-ASCII characters are similarly quoted by git (entire path wrapped in `"..."` with octal escapes inside), and the newline-based parser treats the quoted string as a literal path.

3. `src/workspace/git-history.ts` (line 412) — `git show --patch` output contains escaped paths in `diff --git a/... b/...` headers. The regex in `parseCommitPatchEntries()` (line 381) does not handle quoted/escaped paths.

4. No octal escape decoding utility exists anywhere in the codebase.

**Note:** `git diff-tree` commands in `git-history.ts` (lines 410-411) correctly use the `-z` flag and are not affected.

### Steps to reproduce

1. Create a git repository with a file or directory whose name contains non-ASCII characters (e.g., `docs/提出書類/spec.md`)
2. Make a commit that adds or modifies files under that path
3. Open the Kanban web UI and navigate to the diff/changes view for that commit or working copy
4. Observe that the directory name is displayed as octal escape sequences (e.g., `\345\217\220...`) instead of the actual characters

### Expected behavior

Non-ASCII file and directory names should be displayed correctly as readable text in the diff view (e.g., `提出書類` instead of `\345\217\220\345\207\272\346\233\270\351\241\236`).

### Kanban version

0.1.52

### Environment

Windows 11 (WSL2 Linux 6.6.87.2), Node 25.0.0, npm 11.6.2

hatai:
## Summary

- Inject `-c core.quotepath=false` into `runGit()` so all git commands emit raw UTF-8 paths instead of octal escape sequences for non-ASCII characters
- Apply the same flag to the two direct `execFile("git", ...)` calls in `search-workspace-files.ts` that bypass `runGit`
- Add a test case verifying Japanese file paths are returned as readable UTF-8

Fixes #146

## Before / After

| Before | After |
|--------|-------|
| Arc_34A6P2wZG8 | Arc_J5HUSHdVqu_mosaic |

## Changes

| File | Change |
|------|--------|
| `src/workspace/git-utils.ts` | Prepend `-c core.quotepath=false` to args in `runGit()` |
| `src/workspace/search-workspace-files.ts` | Add same flag to 2 direct `execFile` calls |
| `test/runtime/git-history.test.ts` | New test: commit with Japanese path returns UTF-8 |
| `test/runtime/task-worktree.test.ts` | Update mock helper to transparently skip `-c` flags |

## Test plan

- [x] Existing test suite passes (329/329)
- [x] New non-ASCII path test passes
- [x] Manual verification: Kanban web UI displays Japanese directory names correctly in diff view

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -126,3 +126,26 @@ jobs:
 
             ${{ steps.compare.outputs.link }}
           prerelease: ${{ contains(steps.resolve_tag.outputs.tag, '-') }}
+
+      - name: Post release to Slack
+        uses: slackapi/slack-github-action@v3.0.1
+        with:
+          method: chat.postMessage
+          token: ${{ secrets.SLACK_RELEASE_BOT_TOKEN }}
+          payload: |
+            channel: "C0APVKGGZFC"
+            text: "Kanban ${{ steps.resolve_tag.outputs.tag }}"
+            blocks:
+              - type: "section"
+                text:
+                  type: "mrkdwn"
+                  text: "*Kanban ${{ steps.resolve_tag.outputs.tag }}*"
+              - type: "divider"
+              - type: "section"
+                text:
+                  type: "mrkdwn"
+                  text: ${{ toJSON(steps.changelog.outputs.body) }}
+              - type: "context"
+                elements:
+                  - type: "mrkdwn"
+                    text: "<https://github.com/${{ github.repository }}/releases/tag/${{ steps.resolve_tag.outputs.tag }}|Full Changelog>: ${{ steps.compare.outputs.link }}"
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## [0.1.54]
+
+- Task agent chat panel resizing now persists when navigating between tasks
+
 ## [0.1.53]
 
 - Added `/clear` slash command to reset the Cline agent chat session
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "kanban",
-  "version": "0.1.53",
+  "version": "0.1.54",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "kanban",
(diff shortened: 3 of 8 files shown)


saoudrizwan:
I looked into this in detail because I was also surprised this is not the Git default.

From what I can tell, Git keeps `core.quotepath=true` by default mainly for backward compatibility with older terminals and scripts that expect escaped ASCII-safe output. For app/runtime usage like Kanban, overriding per command with `-c core.quotepath=false` makes sense since we want readable UTF-8 paths.

I checked out this PR locally, tested it myself, and did not see any regressions for existing users. Non-ASCII paths render correctly, and normal ASCII workflows still behave as expected.

I also added a runtime regression test for `searchWorkspaceFiles` that covers both tracked and untracked non-ASCII paths, so we keep this behavior protected going forward.

Thanks for the contribution, this is a great fix.

PR #189, opened 2026-04-02
fix(featurebase): open feedback widget from settings

johnwschoi:
## Summary
Fix the `Share Feedback` button in Settings so the Featurebase modal opens reliably, closes the Settings dialog only after the widget actually opens, and falls back to the GitHub links when the draft provider is not managed Cline OAuth.

## Root cause
There were two separate issues in the Settings flow:

1. The Settings dialog could unmount the `Share Feedback` trigger before the Featurebase SDK finished handling the click.
2. The Settings footer was based on persisted provider settings instead of the live draft provider selection inside the dialog.

That combination could leave the button visible but non-functional, and could also show the wrong footer action while the user was changing providers in Settings.

## What changed
- keep the visible `Share Feedback` button as the real SDK-owned `data-featurebase-feedback` trigger
- initialize the Featurebase feedback widget with an SDK callback and track `widgetOpened` events in the hook
- close the Settings dialog only after the SDK reports that the widget opened
- support clicking the Featurebase backdrop to close the feedback modal through the SDK-supported `closeWidget` message
- make the Settings footer follow the live in-dialog Cline provider draft instead of only persisted config
- keep the GitHub `Report Issue` / `Feature Request` fallback when the selected draft provider is not managed Cline OAuth
- align `vitest.config.ts` with the app aliases/defines needed to import the real Settings dialog in tests

## Verification
- manually verified on the PR worktree build at `http://127.0.0.1:3490/kanban-share-feedback-open`
- `npm --prefix web-ui run test -- src/runtime/native-agent.test.ts src/components/runtime-settings-dialog.test.tsx src/components/featurebase-feedback-button.test.tsx src/hooks/use-featurebase-feedback-widget.test.tsx`
- `npm --prefix web-ui run typecheck`
- `npm run build`

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -101,6 +101,12 @@ jobs:
           POSTHOG_KEY: ${{ secrets.POSTHOG_KEY }}
           POSTHOG_HOST: ${{ secrets.POSTHOG_HOST }}
           SENTRY_AUTH_TOKEN: ${{ secrets.SENTRY_AUTH_TOKEN }}
+          OTEL_TELEMETRY_ENABLED: ${{ secrets.OTEL_TELEMETRY_ENABLED }}
+          OTEL_LOGS_EXPORTER: otlp
+          OTEL_METRICS_EXPORTER: otlp
+          OTEL_EXPORTER_OTLP_PROTOCOL: ${{ secrets.OTEL_EXPORTER_OTLP_PROTOCOL }}
+          OTEL_EXPORTER_OTLP_ENDPOINT: ${{ secrets.OTEL_EXPORTER_OTLP_ENDPOINT }}
+          OTEL_EXPORTER_OTLP_HEADERS: ${{ secrets.OTEL_EXPORTER_OTLP_HEADERS }}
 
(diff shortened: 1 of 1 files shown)


johnwschoi:
@greptileai

web-ui/vitest.config.ts
```diff
@@ -1,15 +1,22 @@
...
 import react from "@vitejs/plugin-react";
 import { defineConfig } from "vitest/config";
 
+const rootPkg = JSON.parse(readFileSync(resolve(__dirname, "../package.json"), "utf-8")) as { version: string };
+
 export default defineConfig({
 	plugins: [react()],
+	define: {
+		__APP_VERSION__: JSON.stringify(rootPkg.version),
+	},
```
johnwschoi, line 12: /tmp/kanban-share-feedback-open/web-ui/src/components/runtime-settings-
  dialog.test.tsx imports the real /tmp/kanban-share-feedback-open/web-ui/src/
  components/runtime-settings-dialog.tsx, and that component depends on build-time
  config that Vitest did not previously provide

PR #204, opened 2026-04-03
allow https flags for kanban and passcode auth

maxpaulus43:
image

### DIFF
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -527,9 +527,7 @@ async function runMainCommand(options: CliOptions, shouldAutoOpenBrowser: boolea
 			const passcode = generatePasscode();
 			generateInternalToken();
 			// NOTE: passcode is printed ONLY here and never stored in logs or env.
-			console.log(
-				`\n🔐 Remote access passcode: ${passcode}\n\nShare this with users who need access. It expires after 24h.\n`,
-			);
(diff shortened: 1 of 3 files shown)


scripts/dogfood.mjs
```diff
@@ -367,12 +421,24 @@ async function main() {
...
+			launchArgs.push("--host", args.host);
+		}
 		if (skipShutdownCleanup) {
 			launchArgs.push("--skip-shutdown-cleanup");
 		}
 		if (args.noOpen) {
 			launchArgs.push("--no-open");
 		}
+		if (args.https) {
+			launchArgs.push("--https");
```
maxpaulus43, line 434: https is not necessary for remote kanban, but browsers hate visiting regular http websites. This will improve the experience for safari and make it much easier to turn kanban into a PWA (which requires https)

src/commands/hooks.ts
```diff
@@ -400,6 +400,10 @@ async function ingestHookEvent(args: HooksIngestArgs): Promise<void> {
 			httpBatchLink({
 				url: buildKanbanRuntimeUrl("/api/trpc"),
 				maxItems: 1,
+				fetch: async (url, options) => {
```
maxpaulus43, line 403: we customize fetch only when https flag is set (so that kanban can trust local certificates if present)

PR #223, opened 2026-04-04
feat(agents): add Pi runtime agent support

raunak42:
_The contribution guide suggests opening an issue first for non-trivial agent integrations. I’m opening this PR directly because this is a focused, test-covered Pi agent integration, but I’m happy to adapt this into an issue-first flow or split the work differently if that’s preferred._

## What is Pi?
Pi is a minimal terminal coding harness with interactive, JSON/RPC, and SDK integration modes.

Links:
- Project/docs: https://github.com/badlogic/pi-mono
- npm package: https://www.npmjs.com/package/@mariozechner/pi-coding-agent
- Website: https://pi.dev

Relevant Pi traits for this integration:
- it has a real interactive terminal UI
- it exposes structured integration surfaces like RPC mode and session logs
- its core semantics differ from some other agents: no built-in plan mode, no permission popups, and provider/login/setup can matter even when the binary is installed

## Why Pi?
- Pi is noticeably faster than Codex even with the same underlying models, codex CLI suffers from bloat and the speed keeps degrading the more you use it in a single session
- it is simpler and less bloated than many other agent CLIs
- it already provides useful integration primitives, especially RPC mode and session logs
- it fits Kanban’s terminal-agent model well

### 1. Pi as a selectable runtime agent
This PR adds Pi to Kanban's runtime-agent flow.

It includes:
- runtime agent catalog/config support
- runtime selection surfaces like settings and onboarding
- supported-agent prompt/system-prompt plumbing

Why it matters:
- Pi becomes a real selectable Kanban runtime agent
- the rest of the app can treat Pi like the other supported runtime agents

image

image

### 2. Real Pi terminal/TUI support
Pi task sessions run in the real PTY/TUI.

It includes:
- Pi launching through a wrapper that preserves the live terminal
- Kanban tracking progress separately from that live terminal session

Why it matters:
- Pi is surfaced the same way as Kanban's other terminal agents
- login, setup, and recovery happen in the same live terminal session the user is already using

image

image

### 3. Session-log-based task-state syncing
Kanban syncs Pi task state from Pi session logs.

It includes:
- per-task Pi session directories
- watching Pi session `.jsonl` logs
- mapping Pi entries into activity, in-progress, and review transitions
- clean-exit reconciliation so tasks are less likely to get stranded

Why it matters:
- Pi does not expose the same structured hook behavior as some other agents during interactive setup/login flows
- Pi session logs are the most reliable structured source of truth for progress and completion

### 4. Readiness handling for “installed but not ready” Pi sessions
Kanban performs a lightweight Pi RPC readiness probe before starting Pi task sessions.

It includes:
- probing Pi readiness through RPC
- allowing the session to start even when login/provider/model setup is still required
- treating readiness failures as recoverable session state instead of hard start failures

Why it matters:
- for Pi, `installed on PATH` is not the same thing as `ready to work on a task`
- keeping the session open preserves the most natural recovery path: finishing setup in the live terminal

### 5. Recovery UX for login/setup-blocked Pi sessions
Pi support includes explicit recovery UX for blocked login/setup states.

It includes:
- warning state plus a dedicated manual-resume flag
- a board-card user-action-needed state instead of misleading `Thinking...`
- terminal recovery guidance plus a manual `Resume task` action
- backend-driven resend of the original task prompt once setup is complete

Why it matters:
- blocked Pi sessions are visible in the board instead of looking like normal agent progress
- manual resend is more reliable than guessing when the live Pi terminal is ready for automatic replay

image

image

### 6. Unsupported Pi controls are disabled
Pi-specific unsupported controls are disabled in the UI.

It includes:
- disabled Pi plan mode in task creation/edit flows
- disabled Pi bypass-permissions/autonomous UI in runtime settings
- Pi-specific explanatory copy for those disabled controls

Why it matters:
- Pi core does not provide equivalent built-in semantics for these controls
- disabling unsupported behavior is clearer and more maintainable than approximating mismatched behavior

image

## How this follows the agent integration guidance

The contribution guide says a good new CLI agent integration PR should include agent detection, session startup/lifecycle handling, prompt integration, terminal/hook support, and tests. This PR covers those areas:

### Agent detection
- Pi was added to the runtime agent catalog/config flow
- Pi can be selected in runtime settings and onboarding
- Pi participates in installed/runnable runtime-agent handling

### Session startup and lifecycle management
- Pi launches as a real PTY session
- Pi uses per-task session directories
- Pi disables auto-restart to avoid relaunch loops
- Pi integrates with session-manager warning state and exit reconciliation

### Prompt integration
- Pi uses the same supported-agent prompt/system-prompt flow as the other runtime agents
- the recovery path resends the original task prompt through Kanban’s backend session-input path

### Terminal integration and hook/state support
- Pi keeps the live terminal visible
- Kanban tracks Pi state through a wrapper plus session-log watching
- Pi task progress and completion feed back into Kanban task state

### Tests
- added Pi-specific runtime tests for wrapper/watcher behavior, readiness probing, adapter launch behavior, and session-manager behavior
- added Pi-specific web tests for board-card state, terminal recovery UI, runtime settings behavior, and task editor behavior

## Key decisions / tradeoffs
- Pi runs in the real PTY/TUI
- Pi session logs are the main structured source of truth for task-state syncing
- readiness failures do not block task start; recovery happens in the live terminal
- prompt replay is manual because Kanban cannot reliably detect when the live Pi terminal is ready for input again
- unsupported Pi semantics are disabled rather than approximated with mismatched behavior

## Validation
Ran locally:
- `npm test`
- `npm --prefix web-ui run test`
- `npm run check`
- `npm run build`

### DIFF
--- a/src/commands/pi-hooks.ts
+++ b/src/commands/pi-hooks.ts
@@ -33,6 +33,11 @@ export function buildPiWrapperChildArgs(agentArgs: string[], sessionDir: string)
 	const hasSessionDir = childArgs.some((arg) => {
 		return arg === "--session-dir" || arg.startsWith("--session-dir=");
 	});
+	if (!hasSessionDir) {
+		childArgs.unshift(sessionDir);
+		childArgs.unshift("--session-dir");
+	}
+	return childArgs;
 }
 
 export async function startPiSessionWatcher(
--- a/src/trpc/runtime-api.ts
+++ b/src/trpc/runtime-api.ts
@@ -368,13 +368,6 @@ export function createRuntimeApi(deps: CreateRuntimeApiDependencies): RuntimeTrp
 						summary = terminalManager.applyNeedsManualPromptResend(body.taskId, false) ?? summary;
 					}
 				}
-				if (!summary) {
-					return {
-						ok: false,
-						summary: null,
-						error: "Task session is not running.",
-					};
-				}
 				return {
 					ok: true,
 					summary,
--- a/test/runtime/terminal/pi-readiness.test.ts
+++ b/test/runtime/terminal/pi-readiness.test.ts
@@ -63,6 +63,38 @@ describe("probePiReadiness", () => {
 		);
 	});
 
+	it("treats a successful prompt response as ready even when Pi writes benign stderr output", async () => {
+		const child = new FakeChildProcess();
+		const spawnMock = vi.fn(() => child as never);
+
+		const probePromise = probePiReadiness(
+			{
+				binary: "pi",
+				args: [],
+				cwd: "/tmp/worktree",
+			},
+			{ spawn: spawnMock },
+		);
+
+		child.stderr.write("Loading provider metadata...\n");
+		child.stdout.write(
+			`${JSON.stringify({
+				id: "kanban-pi-readiness",
+				type: "response",
+				command: "prompt",
+				success: true,
+			})}\n`,
+		);
+		child.emit("exit", 0, null);
+
+		await expect(probePromise).resolves.toEqual({
+			status: "ready",
+			reason: null,
+			message: null,
+			rawMessage: null,
+		});
+	});
+
 	it("maps stderr startup failures to not ready when Pi exits before a prompt response", async () => {
 		const child = new FakeChildProcess();
 		const spawnMock = vi.fn(() => child as never);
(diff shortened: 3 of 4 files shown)


raunak42:
Thanks, that makes sense. I’m aligning kanban's pi integration with how the existing CLI agents actually behave, rather than introducing pi-only "user action" oriented ux.
 
I also found a pi-side bug: when pi starts with an initial prompt and login is required, the initial prompt gets lost because prompt submission fails before it’s retained anywhere. 
 
 I opened an upstream pi issue for that and already have a local fix. If that lands, pi should behave much closer to codex 
 here without kanban-specific recovery UI.

PR #241, opened 2026-04-06
max/kanban custom model per task

maxpaulus43:
- **feat: add per-task agent/model/provider override fields**
- **feat(cli): add --agent-id, --cline-provider, --cline-model flags to task create and update commands**
- **feat: wire per-task agent/model overrides into runtime session launch**
- **feat(web-ui): add per-task agent and Cline model/provider selector UI**
- **fix: preserve task model fields on restore-from-trash and in detail view model picker**

https://github.com/user-attachments/assets/29724e64-4cb0-429a-a806-439a6988829d


maxpaulus43:
@greptileai

web-ui/src/state/board-state.ts
johnwschoi, line 146: @maxpaulus43  do we have to add agentId, clineProviderId, clineModelId here to preserve those fields when hydrating board data? lets say for cases like reload?

src/trpc/runtime-api.ts
```diff
@@ -171,7 +171,7 @@ export function createRuntimeApi(deps: CreateRuntimeApiDependencies): RuntimeTrp
 				const previousTerminalAgentId = body.resumeFromTrash
 					? (terminalManager.getSummary(body.taskId)?.agentId ?? null)
 					: null;
-				const effectiveAgentId = previousTerminalAgentId ?? scopedRuntimeConfig.selectedAgentId;
+				const effectiveAgentId = previousTerminalAgentId ?? body.agentId ?? scopedRuntimeConfig.selectedAgentId;
 				let useClinePath = effectiveAgentId === "cline";
```
johnwschoi, line 190: Can we document whether the request body or the persisted task card is the canonical source of truth for per-task config if they ever diverge? or would that not be possible?

src/trpc/runtime-api.ts
```diff
@@ -171,7 +171,7 @@ export function createRuntimeApi(deps: CreateRuntimeApiDependencies): RuntimeTrp
...
 				const shouldProbePersistedClineSession =
 					body.resumeFromTrash && !useClinePath && previousTerminalAgentId === null;
```
johnwschoi, line 192: Should `resumeFromTrash` ignore current card-level provider/model overrides so a resumed Cline task preserves its original launch configuration the same way agent precedence already preserves the prior session?

src/commands/task.ts
```diff
@@ -970,6 +1018,9 @@ export function registerTaskCommand(program: Command): void {
 		.option("--start-in-plan-mode [value]", "Set plan mode (true|false). Flag-only implies true.")
 		.option("--auto-review-enabled [value]", "Enable auto-review behavior (true|false). Flag-only implies true.")
 		.option("--auto-review-mode <mode>", "Auto-review mode: commit | pr | move_to_trash.", parseAutoReviewMode)
+		.option("--agent-id <id>", "Agent override: cline | claude | codex | droid | gemini | opencode.")
+		.option("--cline-provider <id>", "Cline provider override (e.g. anthropic, openai, cline).")
+		.option("--cline-model <id>", "Cline model override (e.g. claude-sonnet-4-20250514).")
```
johnwschoi: nit: nice, do we need an option to clear the override? like way to goback to default behavior?

src/trpc/runtime-api.ts
maxpaulus43: If I set the model for a task to gemini and then delete it and then restore it, i expect gemini to still be the model after restoration

web-ui/src/state/board-state.ts
maxpaulus43: yes good catch.

src/trpc/runtime-api.ts
maxpaulus43: added a comment

src/commands/task.ts
maxpaulus43: added a "default" value

johnwschoi (dismissed):
lgtm, most of my comments are addressed.

saoudrizwan:
A few things stood out while reviewing this end to end:

1. `web-ui/src/hooks/use-task-editor.ts:377`
The multi-create flow drops the new per-task override fields. `handleCreateTask()` passes `agentId`, `clineProviderId`, and `clineModelId`, but `handleCreateTasks()` does not. So if someone uses the new override UI and then uses the "split into tasks" path, every created task silently falls back to the workspace default.

2. `web-ui/src/components/task-agent-model-picker.tsx:128` and `src/cline-sdk/cline-provider-service.ts:554`
The task-level provider picker looks effectively unusable right now. The picker filters to `p.enabled`, but `getProviderCatalog()` only marks the currently selected global provider as enabled. That means the override dropdown can only show the current default provider, not the other providers you might actually want to override to.

3. `web-ui/src/components/detail-panels/cline-agent-chat-panel.tsx:315`, `web-ui/src/hooks/use-runtime-settings-cline-controller.ts:462`, and `src/cline-sdk/cline-provider-service.ts:704`
Changing the model from the detail view still appears to mutate the global provider settings, not just the current task. `persistClineModelSettings()` calls `saveProviderSettings()`, and that writes to the shared provider settings for that provider. Then `onClineModelChanged()` also writes the explicit override onto the current task card.

That last point is the one I think we need to be especially careful about, because the current data model seems to have two layers:

1. A global provider-level default, stored in the shared Cline provider settings keyed by provider id.
2. Optional task-level overrides on the card: `agentId`, `clineProviderId`, and `clineModelId`.

My read of the current behavior is:

- If a task has no override fields set, it inherits from the global provider settings.
- If a task has `clineProviderId` / `clineModelId` set, that task is pinned to those values.
- The detail-view model picker currently does both: it updates the current task's override fields and also saves the provider's global default model through `saveProviderSettings()`.

So the practical effect is not "just this task changed models". It looks more like:

- This task gets an explicit override.
- Any other task that is still inheriting the default for that same provider will now see the new global model.
- Tasks that already have their own explicit override should stay on their own model.

Also, `saveClineProviderSettings(workspaceId, ...)` is called with a workspace id, but the runtime route ignores workspace scope and writes through the shared provider service anyway, so this seems global rather than workspace-local.

I think it would help to make the intended ownership model very explicit before merge: when the user changes the model in the task detail view, should that be changing only that task, or should it also update the shared default for future tasks and other cards that inherit defaults?

I believe it should update the shared default for future tasks and other cards that inherit defaults, and want to confirm that things are implemented this way.

PR #248, opened 2026-04-07
feat: add editable task titles and sync session metadata

robinnewhouse:
## Summary
- add first-class editable task titles across the Kanban UI, task create/edit flows, persisted board state, and CLI/runtime contracts
- keep the Kanban board card title as the canonical source of truth, while forwarding a normalized `taskTitle` copy into Cline session metadata when a task session starts
- preserve backward compatibility for existing boards by deriving titles from prompts when persisted cards do not yet have an explicit title

## Why
PR #44 established the intended UX for editable task titles, but it had gone stale relative to `main`. This branch rebases that intent onto the current architecture, finishes the end-to-end wiring, and keeps the title model simple:

- tasks exist before sessions, so the board owns the canonical title
- sessions may receive a best-effort metadata copy for convenience
- the runtime does not read titles back from session metadata, avoiding a second source of truth and extra coupling

## What changed
- add stored `title` support for runtime board cards, with prompt-derived fallback defaults for older board data
- add shared task-title helpers for derivation and normalization
- add `--title` support to CLI task create/update flows and document it in prompts/contracts
- update board card rendering to display `card.title` directly instead of width-based prompt splitting
- add inline title editing on board cards and wire title editing through task create, inline create, and edit flows
- persist title changes through board-state mutations and keep UI state cleanup coherent across save/cancel/reset flows
- thread optional `taskTitle` through runtime session start so SDK sessions receive a normalized title copy on start
- add focused tests for title derivation, board persistence, UI editing flows, runtime API wiring, and best-effort SDK session metadata sync behavior

## Architectural notes
- `BoardCard.title` is the source of truth
- `prompt` remains the full task instruction body
- `taskTitle` on session start is a one-way propagation only
- no runtime/session read path was introduced for task titles
- older persisted boards are normalized on read so existing data remains valid

## Review notes
- the implementation follows the prior commit history from the stale PR, but trims dead width-measurement title logic and reduces coupling by moving title-save behavior into board/task state helpers
- automated review feedback did not identify any P0/P1 issues; the remaining comments were robustness/style observations around defensive normalization and inline edit cancellation behavior
- the current implementation keeps those trade-offs explicit while preserving the simpler single-source-of-truth model

## Screenshots

Uploading Screen Recording 2026-04-06 at 8.55.43 PM.mov…

## Test plan
- [x] `npm run build`
- [x] `npm run lint`
- [x] `npm run typecheck`
- [x] `npm test`
- [x] `npm --prefix web-ui run test`


shey-cline (dismissed):
Tested and LGTM!

robinnewhouse:
Screenshot 2026-04-09 at 3 56 02 PM
Screenshot 2026-04-09 at 3 56 22 PM

shey-cline (approved):

PR #259, opened 2026-04-07
fix: replace Kanban-side credit limit workarounds with SDK-native handling and board-level credit banner

johnwschoi:
## Summary

Replaces fragile Kanban-side credit limit workarounds with SDK-native handling, adds board-level and card-level UI for out-of-credits state, and scopes credit-limit detection to the Cline provider only.

From the board view with using kanban agent on the left
Screenshot 2026-04-08 at 5 17 56 PM

Window inside the the task (note: moving from trash to review starts clean session so Out of Credits will show after they attempt to talk. if not we would have to update the architecture further to persist latest hook activity.
Screenshot 2026-04-08 at 5 30 13 PM

### Changes

- **SDK-native error handling**: Trust the SDK's `recoverable` flag instead of overriding it Kanban-side. Remove the old `shouldForceAbortFromTaskEvent` retry-loop workaround.
- **Shared `isCreditLimitError` helper**: Consolidated duplicate detection logic into a single helper in `cline-session-state.ts` with comprehensive string patterns.
- **Board-level credit banner**: New `CreditLimitBanner` component shown above columns when any task hits a credit limit. Dismissable, auto-reappears on new incidents.
- **Card-level indicator**: Orange triangle icon + "Out of credits" text on affected cards via `isCardCreditLimitError` helper.
- **`interrupted` state support**: Banner and card badge persist after abort transitions the session to `interrupted`, since `notificationType` carries forward.
- **Cline-provider scoping**: Credit-limit detection only fires for the Cline provider (`providerId === "cline"`). Non-Cline providers (custom OpenAI keys, etc.) get standard error handling — consistent with how Kanban handles all other error types.
- **Tightened detection**: Narrowed "payment required" to "402 payment required" to avoid false positives. Added `message?` field to `ClineSdkErrorEvent` type to remove unsafe double cast.
- **Reusable `Link` component**: Extracted shared `Link` UI component with `external` prop for consistent external link handling.
- **`agentEvent.message` fallback**: Credit-limit detection falls back to `agentEvent.message` when `error` is absent, covering SDK edge cases.

### Test coverage

- Unit tests for credit-limit detection on start/send failure, agent event abort, recovery notice suppression, `agentEvent.message` fallback, and non-Cline provider bypass.
- Banner component tests for render/dismiss/reshow cycle and state gating.
- E2E verified locally against core-platform with $0 org balance.

## Test plan

- [x] Unit tests pass (57 backend + 4 banner)
- [x] Credit-limit banner appears on board when Cline provider returns 402
- [x] Card shows orange triangle + "Out of credits" 
- [x] Both indicators persist after abort (interrupted state)
- [x] Dismiss button works, banner reappears on new credit-limit event
- [x] Non-Cline providers show standard error handling, no credit-limit UI

### DIFF
--- a/src/cline-sdk/cline-message-repository.ts
+++ b/src/cline-sdk/cline-message-repository.ts
@@ -21,6 +21,7 @@ export interface ClineMessageRepository {
 	onSummary(listener: (summary: RuntimeTaskSessionSummary) => void): () => void;
 	onMessage(listener: (taskId: string, message: ClineTaskMessage) => void): () => void;
 	setTaskEntry(taskId: string, entry: ClineTaskSessionEntry): void;
+	clearHydratedTaskMessages(taskId: string): void;
 	getTaskEntry(taskId: string): ClineTaskSessionEntry | null;
 	getSummary(taskId: string): RuntimeTaskSessionSummary | null;
 	listSummaries(): RuntimeTaskSessionSummary[];
@@ -58,6 +59,10 @@ export class InMemoryClineMessageRepository implements ClineMessageRepository {
 
 	setTaskEntry(taskId: string, entry: ClineTaskSessionEntry): void {
 		this.entries.set(taskId, entry);
+		this.hydratedMessagesByTaskId.delete(taskId);
+	}
+
+	clearHydratedTaskMessages(taskId: string): void {
 		this.hydratedMessagesByTaskId.delete(taskId);
 	}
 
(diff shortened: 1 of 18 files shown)


johnwschoi:
per greptile comment   1. The early-return guard was not changed — main has the exact same guard
  without !request.resumeFromTrash. This PR didn't touch it.
  2. No trash-restore test was deleted — the diff shows zero removed lines
  mentioning "trash" or "resume" in the test file.

PR #261, opened 2026-04-07
max/remote file browser

maxpaulus43:
- **feat: add server-side directory listing API endpoint for remote file browsing**
- **feat: add RemoteFileBrowserDialog component for remote filesystem browsing**
- **feat: add git clone support in backend addProject flow**
- **feat: unified Add Project dialog with server path, browse, and git clone modes**

Tests:
- tested by hosting kanban on my mac and exposing via ngrok, then accessing via my phone and via my windows computer
- hosted kanban via a docker container exposing port 3484 and accessing via my mac browser
- hosted kanban on my wsl workspace and accessing via my windows browser

https://github.com/user-attachments/assets/916200e2-bcc5-4eef-b3ec-6c81dcfd9a58


saoudrizwan (commented):
High: the new path handling is still POSIX-only, so the add/browse/clone flow breaks on a native Windows runtime.

I pulled this locally and walked through the path flow end to end. The core problem is that several places are still treating server paths as if they always use `/`, both in the runtime sandbox checks and in the browser-side path manipulation.

Where this shows up

- `src/trpc/projects-api.ts`
  - `listDirectoryContents()` uses `rootPath.endsWith("/") ? rootPath : `${rootPath}/`` and then `startsWith(normalizedRoot)` to decide whether an absolute path is inside the sandbox.
- `src/workspace/git-clone.ts`
  - `validateCloneDestination()` uses the same `serverCwd + "/"` prefix strategy.
- `web-ui/src/components/add-project-dialog.tsx`
  - builds absolute paths with string concatenation using `/`.
- `web-ui/src/components/directory-autocomplete.tsx`
  - strips the server root with a slash-based prefix assumption.
- `web-ui/src/components/remote-file-browser-dialog.tsx`
  - breadcrumbs split on `/` only.

Why this is a real bug

On a Windows runtime, the server will return paths like:

```text
C:\workspace
C:\workspace\repo-a
```

But the current checks normalize the root into something like:

```text
C:\workspace/
```

That means a valid in-root path such as `C:\workspace\repo-a` no longer matches the prefix check, even though it is clearly inside the root. The same issue shows up again in the web UI when it tries to turn absolute server paths back into root-relative values for autocomplete and breadcrumb navigation.

So the failure mode is not hypothetical. On Windows hosts, you can end up with behavior like:

- browsing into a subdirectory fails even though the path came from the server itself
- clone destinations inside the server root are rejected as if they were outside the sandbox
- autocomplete selections cannot be converted back into the expected UI-relative path
- breadcrumb rendering becomes incorrect because `split("/")` never breaks a `C:\...` path into segments

This is especially likely to show up when Kanban is running on a remote Windows machine, but it is not limited to remote servers. Anyone running Kanban on Windows can hit it.

Suggested direction

I would strongly recommend replacing the string-prefix sandbox checks with a helper based on `path.resolve()` plus `path.relative()`. That approach is separator-safe and works correctly on macOS, Linux, and Windows.

Something along these lines for the runtime side:

```ts
import * as path from "node:path";

function isPathWithinRoot(rootPath: string, candidatePath: string): boolean {
  const resolvedRoot = path.resolve(rootPath);
  const resolvedCandidate = path.resolve(candidatePath);
  const relativePath = path.relative(resolvedRoot, resolvedCandidate);

  if (relativePath === "") {
    return true;
  }

  return relativePath !== ".." && !relativePath.startsWith(`..${path.sep}`) && !path.isAbsolute(relativePath);
}
```

Then use that helper in both places:

- `projects-api.ts` for `listDirectoryContents()`
- `git-clone.ts` for `validateCloneDestination()`

For the browser side, I would avoid hand-building server paths with `"/"`. The UI does not need to guess the host OS, but it does need to respect the separator used by the paths the server is already returning.

A practical approach would be:

- keep the text field UI root-relative if you want, for example `/repo-a/src`
- add a small helper that converts `rootPath + uiRelativePath -> absolute server path`
- add a second helper that converts `absolute server path -> uiRelativePath`
- detect the separator from the server path itself, for example `\\` vs `/`
- build breadcrumbs by splitting on both separators, for example `/[\\/]+/`

Very roughly:

```ts
function getServerSeparator(pathValue: string): "/" | "\\" {
  return pathValue.includes("\\") ? "\\" : "/";
}

function toServerAbsolute(rootPath: string, uiPath: string): string {
  const sep = getServerSeparator(rootPath);
  const parts = uiPath.split(/[\\/]+/).filter(Boolean);
  return [rootPath.replace(/[\\/]+$/, ""), ...parts].join(sep);
}
```

I would also add explicit Windows-style tests using `node:path.win32` for the sandbox helper and a couple of UI helper tests for:

- `C:\workspace\repo` is accepted inside `C:\workspace`
- `C:\workspace-other\repo` is rejected
- absolute server entry paths can be converted back to the expected UI-relative value
- breadcrumbs render correctly for `C:\workspace\repo\src`

I think the feature itself makes sense, but I would treat this as blocking because the current implementation bakes in slash-only assumptions in several layers of the flow.

saoudrizwan:
Pushed commit 41246bb with:

- Removed the in-app picker sandbox to process.cwd by switching directory browsing root to filesystem root (resolve(serverCwd, '/')).
- Updated clone destination validation to support a broader root boundary so clone paths can be outside launch cwd when called from projects API.
- Added and updated runtime tests for browsing outside launch cwd, filesystem-root traversal behavior, and broader clone destination support.
- Removed the 'Server root' label from the Add Project dialog since the picker now uses filesystem root.

Why:
- The previous cwd-bound sandbox blocked valid container paths... when Kanban was launched from a different cwd.
- This aligns Add Project behavior with the expectation that users can pick any directory visible inside the container filesystem.

PR #262, opened 2026-04-07
fix: clear sidebar agent chat UI when /clear command is used

robinnewhouse:
## Problem

When a user types `/clear` in the kanban sidebar agent chat, the backend correctly clears the session and broadcasts a `task_chat_cleared` WebSocket event, but the chat UI does not update — old messages remain visible.

## Root Cause

The `incomingMessages` effect in `useClineChatSession` (`web-ui/src/hooks/use-cline-chat-session.ts`) treated empty arrays the same as `null`:

```ts
if (!incomingMessages || incomingMessages.length === 0) {
    return; // never clears local state
}
```

When `task_chat_cleared` fires, the reducer removes the task from `taskChatMessagesByTaskId`, so `incomingMessages` becomes `[]`. But the effect ignored it.

## Fix

Add a `hasReceivedIncomingMessages` ref to distinguish:
- **"no streamed messages have ever arrived"** (`[]` on initial render) → ignore, preserve loaded history
- **"messages were cleared after previously existing"** (transition from non-empty to `[]`) → reset local state

The ref is reset on task switches.

## Tests

Three new tests in `use-cline-chat-session.test.tsx`:
1. `/clear` send path clears messages via `onLoadMessages` fallback
2. `incomingMessages` populated→empty transition clears local messages
3. Loaded history is preserved when `incomingMessages` is empty-by-default across re-renders (regression guard)

## Verification

```
npx vitest run src/hooks/use-cline-chat-session.test.tsx           # 10/10 pass
npx vitest run src/hooks/use-cline-chat-panel-controller.test.tsx  # 7/7 pass
npx vitest run src/components/detail-panels/cline-agent-chat-panel.test.tsx  # 21/21 pass
npm run web:typecheck                                              # clean
```

https://github.com/user-attachments/assets/93eefb50-e6a3-4278-9006-058f73c912c7

### DIFF
--- a/src/cline-sdk/cline-message-repository.ts
+++ b/src/cline-sdk/cline-message-repository.ts
@@ -21,6 +21,7 @@ export interface ClineMessageRepository {
 	onSummary(listener: (summary: RuntimeTaskSessionSummary) => void): () => void;
 	onMessage(listener: (taskId: string, message: ClineTaskMessage) => void): () => void;
 	setTaskEntry(taskId: string, entry: ClineTaskSessionEntry): void;
+	clearHydratedTaskMessages(taskId: string): void;
 	getTaskEntry(taskId: string): ClineTaskSessionEntry | null;
 	getSummary(taskId: string): RuntimeTaskSessionSummary | null;
 	listSummaries(): RuntimeTaskSessionSummary[];
@@ -58,6 +59,10 @@ export class InMemoryClineMessageRepository implements ClineMessageRepository {
 
 	setTaskEntry(taskId: string, entry: ClineTaskSessionEntry): void {
 		this.entries.set(taskId, entry);
+		this.hydratedMessagesByTaskId.delete(taskId);
+	}
+
+	clearHydratedTaskMessages(taskId: string): void {
 		this.hydratedMessagesByTaskId.delete(taskId);
 	}
 
--- a/src/cline-sdk/cline-task-session-service.ts
+++ b/src/cline-sdk/cline-task-session-service.ts
@@ -677,6 +677,7 @@ export class InMemoryClineTaskSessionService implements ClineTaskSessionService 
 		const existingEntry = this.messageRepository.getTaskEntry(taskId);
 		this.pendingTurnCancelTaskIds.delete(taskId);
 		await this.sessionRuntime.clearTaskSessions(taskId).catch(() => undefined);
+		this.messageRepository.clearHydratedTaskMessages(taskId);
 		if (!existingEntry) {
 			return null;
 		}
--- a/test/runtime/cline-sdk/cline-message-repository.test.ts
+++ b/test/runtime/cline-sdk/cline-message-repository.test.ts
@@ -131,4 +131,26 @@ describe("InMemoryClineMessageRepository", () => {
 		expect(messages.map((message) => message.content)).toEqual(["Live response"]);
 		expect(loadPersistedSession).not.toHaveBeenCalled();
 	});
+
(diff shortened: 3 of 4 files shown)


web-ui/src/hooks/use-cline-chat-session.ts
```diff
@@ -111,9 +113,17 @@ export function useClineChatSession({
 	}, [onLoadMessages, taskId]);
 
 	useEffect(() => {
-		if (!incomingMessages || incomingMessages.length === 0) {
+		if (!incomingMessages) {
 			return;
 		}
+		if (incomingMessages.length === 0) {
```
BarreiroT, line 119: Hmm, can we use a parameter rather than `incomingMessages.length === 0`? Feels like this is a weird side-effect we will forget about

robinnewhouse: Fair, looking into that

robinnewhouse: I tried an explicit clear parameter/counter, but threading it through the chat prop chain added too much coupling, so I reverted that approach. The current fix keeps it local in `useClineChatSession`: clear only on a populated→empty `incomingMessages` transition, with a regression test ensuring empty default arrays on rerender don’t wipe loaded history.

web-ui/src/hooks/use-cline-chat-session.ts
```diff
@@ -111,9 +113,17 @@ export function useClineChatSession({
...
+			if (hasReceivedIncomingMessages.current) {
+				hasReceivedIncomingMessages.current = false;
```
BarreiroT, line 121: Hmm, I am not quite sure I see the need for `hasReceivedIncomingMessages`. It's kinda like: only set messages to empty if they aren't already empty, but it seems like an unnecessary extra load.
If they're empty, what's the problem with clearing them?

We could do something like: `setMessages(old => { if (!old.length) return old; return []})` if the intention is to not rerender when it's not needed, but even then rerendering an empty message array shouldn't be super expensive

Am I missing something?

robinnewhouse: The ref isn't about the re-render cost. it's needed to distinguish "WebSocket stream cleared messages" from "incomingMessages is just the default []". Without it, a parent re-render would wipe history loaded via onLoadMessages too (see the third new test).

BarreiroT: Hmm, can `incomingMessages` actually be `[]` while running? Looks like `use-cline-chat-session` has to deduplicate them

robinnewhouse: incomingMessages can be [] in normal runtime for both:
- “no stream messages yet” (default fallback), and
- “stream cleared” (task_chat_cleared removes the task key).

Without the ref (or a different sentinel contract), those two states are indistinguishable and a rerender can wipe loaded history.

BarreiroT: So.... I might be misunderstanding something here but it seems to me like we're preventing clearing messages on a case when:
- Messages are already clear (because they haven't loaded yet or the conversation is empty)?
- We want to clear messages?

Why is it a problem?

robinnewhouse:
Cross-reference: #272 is an alternative implementation for the same /clear bug family.

It keeps the backend hydrated-cache eviction fix while simplifying frontend semantics via explicit stream sentinel values (null vs empty array). Please compare both PRs side-by-side for maintainability preference.

PR #263, opened 2026-04-07
feat: add Kiro CLI agent

saoudrizwan:
## Summary

This PR adds Kiro CLI as a first-class runtime agent in Kanban, alongside the existing Claude, Codex, Cline, and Droid options.

It wires Kiro through agent discovery, runtime configuration, task launch preparation, hook ingestion, and the web UI so Kiro sessions can participate in the same task lifecycle as the other agents.

## Problem

Kanban already had a mature path for agent-driven task execution, but Kiro CLI was not available as a selectable runtime agent. That meant users could not launch Kiro from Kanban, use Kiro's hooks to move cards between in progress and review, or see Kiro tool activity and final review messages reflected in the task cards.

The tricky part here was not just adding another agent entry. Kiro's hook payload shape differs from the existing agents, especially around tool input payloads and stop events, so the UI metadata pipeline needed to understand real Kiro hook payloads rather than only idealized payloads.

### Runtime integration

- Added `kiro` to the runtime agent contract and launch-supported catalog
- Added a Kiro catalog entry with `kiro-cli` as the binary, `chat` as the base command, and `--trust-all-tools` for autonomous mode
- Updated installed-agent auto-selection so `kiro-cli` can map back to the logical `kiro` agent id
- Added Kiro-specific Linear MCP setup guidance in the appended system prompt

### Launch and hook wiring

- Added a Kiro session adapter in `src/terminal/agent-session-adapters.ts`
- The adapter writes a managed Kiro agent config at `~/.kiro/agents/kanban.json`
- The managed Kiro agent forwards `agentSpawn`, `userPromptSubmit`, `preToolUse`, `postToolUse`, and `stop` into Kanban's hook ingest command
- Resume from trash maps to `--resume`
- Plan mode maps to `/plan ...`
- Autonomous mode maps to `--trust-all-tools`

### Hook parsing and UI metadata

- Added Kiro transcript source inference in `src/commands/hooks.ts`
- Extracted Kiro-specific hook parsing into `src/commands/hook-events/kiro-hook-events.ts`
- Added tool input summarization logic for Kiro payloads, with preference for richer payload shapes like `preToolUse.input` and `tool_input`
- Parsed Kiro stop hooks using `assistant_response`, which is the real field Kiro emits for the final assistant message
- Updated the board card preview logic so stale tool names do not leak into review-state cards when the latest activity is not actually a tool call
- Preserved canonical Kiro tool names like `fs_write` in the preview label instead of aliasing them

### Code organization

- Moved the Codex, Droid, and Kiro hook helper modules into `src/commands/hook-events/`
- Left `src/commands/hooks.ts` in place as the main command entrypoint and ingest coordinator

## Debugging journey

A lot of the work here was about matching Kiro's real hook payloads instead of assumptions.

Early on, tool previews still showed incomplete labels like `fs_write` or `fs_write(str_replace)` instead of the actual file path. That turned out to be caused by two separate issues:

- The parser was sometimes preferring less useful payload branches over richer ones
- The Kiro payload shape for writes can include both `command` and `path`, and the summarizer initially chose `command` before `path`

To resolve that, I captured real Kiro hook payloads from non-interactive runs and adjusted the summarization logic to prefer file paths when present.

I also found that review cards could show a stale bare tool name if the most recent hook only carried a stale `toolName` field alongside non-tool activity like `Waiting for review`. That is now guarded in the board card label resolver so the UI falls back to the human activity text instead of rendering a misleading tool label.

Finally, Kiro's final message field turned out to be `assistant_response`, not the more generic fields used by other agents, so the stop hook parser now reads that explicitly.

## Gotchas and decisions

- I kept the managed Kiro hook parser separate from the generic hook normalizer to avoid turning `src/commands/hooks.ts` into a pile of agent-specific conditionals
- I did not add a new hook metadata CLI flag just for Kiro tool summaries. The activity text is derived from the payload directly instead
- I kept Kiro tool names canonical in the UI, for example `fs_write`, because aliasing them to friendlier names hid useful detail and was explicitly backed out during iteration
- The managed Kiro agent name is `kanban`, so Kanban writes `~/.kiro/agents/kanban.json`

### Automated validation

Run:

```bash
npm run lint
npm run typecheck
npm run web:typecheck
npm run test
npm run web:test
```

These all pass on this branch.

### Manual testing

1. Install Kiro CLI and make sure `kiro-cli` is on `PATH`
2. Open Kanban runtime settings and confirm Kiro appears as an agent option
3. Start a task with Kiro and verify Kanban launches `kiro-cli chat`
4. Confirm Kiro hook activity moves the card into in progress and then to review
5. Confirm tool previews show canonical tool names with useful details, for example `fs_write(path/to/file)`
6. Confirm review cards show the final assistant message when available instead of only `Waiting for review`

## Risk

The main product caveat is that Kanban currently writes its managed Kiro agent config to `~/.kiro/agents/kanban.json`. If a user already has a personal Kiro agent named `kanban`, this will overwrite that config.


src/prompts/append-system-prompt.ts
```diff
@@ -56,6 +64,8 @@ function renderLinearSetupGuidanceForAgent(agentId: RuntimeAgentId | null): stri
 			return "- If Linear MCP is not available in the current agent (OpenCode), suggest running `opencode mcp add`, then use name `linear` and URL `https://mcp.linear.app/mcp`.";
 		case "droid":
 			return "- If Linear MCP is not available in the current agent (Droid), suggest running: `droid mcp add linear https://mcp.linear.app/mcp --type http`";
+		case "kiro":
+			return "- If Linear MCP is not available in the current agent (Kiro CLI), suggest running: `kiro-cli mcp add linear https://mcp.linear.app/mcp --transport http --scope user`";
```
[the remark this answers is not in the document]
maxpaulus43: kiro cli --scope field only support values: `default`, `workspace`, `global`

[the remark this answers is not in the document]
maxpaulus43:
```bash
  ➜  kanban git:(saoudrizwan/kiro-cli-agent-support) kiro-cli mcp add --help
Add or replace a configured server

Usage: kiro-cli-chat mcp add [OPTIONS]

Options:
      --name <NAME>
          Name for the server (optional when registry is configured - shows interactive menu)

      --scope <SCOPE>
          Scope. This parameter is only meaningful in the absence of agent name

          [possible values: default, workspace, global]

      --command <COMMAND>
          The command used to launch the server (required for stdio servers, not needed for registry or HTTP servers)

      --url <URL>
          The URL for HTTP-based MCP server communication (required for HTTP servers)

      --args <ARGS>
          Arguments to pass to the command. Can be provided as:

          1. Multiple --args flags: --args arg1 --args arg2 --args "arg,with,commas"
          2. Comma-separated with escaping: --args "arg1,arg2,arg,with,commas"
          3. JSON array format: --args '["arg1", "arg2", "arg,with,commas"]'

      --agent <AGENT>
          Where to add the server to. If an agent name is not supplied, the changes shall be made to the global mcp.json

      --env <ENV>
          Environment variables to use when launching the server

      --timeout <TIMEOUT>
          Server launch timeout, in milliseconds

      --disabled
          Whether the server should be disabled (not loaded)

      --force
          Overwrite an existing server with the same name

  -v, --verbose...
          Increase logging verbosity

  -h, --help
          Print help (see a summary with '-h')
```

src/commands/hooks.ts
```diff
@@ -323,6 +327,20 @@ function normalizeHookMetadata(
 	payload: Record<string, unknown> | null,
 	flagMetadata: Partial<RuntimeTaskHookActivity>,
 ): Partial<RuntimeTaskHookActivity> | undefined {
+	const inferredSource = inferHookSourceFromPayload(payload);
```
maxpaulus43, line 295: nit: this sets a precedent to add new conditional statements for every new cli that has different hook payload formats. I could see this turning into a large switch statement here switching on sourceHint?.toLowerCase(), which could get messy. 

Maybe we could use the strategy pattern here and just put something like 
const hookParseStrategy = inferHookStrategyFromPayload(payload)
hookParseStrategy.normalizeHookPayload(payload)

src/core/agent-catalog.ts
```diff
@@ -50,6 +50,14 @@ export const RUNTIME_AGENT_CATALOG: RuntimeAgentCatalogEntry[] = [
 		autonomousArgs: ["--auto", "high"],
 		installUrl: "https://docs.factory.ai/cli/getting-started/quickstart",
 	},
+	{
+		id: "kiro",
```
maxpaulus43, line 54: i'd love a world where all we have to do to add support for a new cli agent is to add something to this file and include a hookPayloadParsing strategy for that cli agent

src/prompts/append-system-prompt.ts
```diff
@@ -56,6 +64,8 @@ function renderLinearSetupGuidanceForAgent(agentId: RuntimeAgentId | null): stri
 			return "- If Linear MCP is not available in the current agent (OpenCode), suggest running `opencode mcp add`, then use name `linear` and URL `https://mcp.linear.app/mcp`.";
 		case "droid":
 			return "- If Linear MCP is not available in the current agent (Droid), suggest running: `droid mcp add linear https://mcp.linear.app/mcp --type http`";
+		case "kiro":
```
maxpaulus43, line 67: these system prompt updates could just exist in the agent-catalog.ts. keeps cli-specific information more consolidated

src/terminal/agent-session-adapters.ts
```diff
@@ -1,4 +1,5 @@
 import { access, readFile } from "node:fs/promises";
```
maxpaulus43, line 1: this file should be split by agent. it's getting too big

web-ui/src/components/runtime-settings-dialog.tsx
```diff
@@ -74,7 +74,7 @@ const GIT_PROMPT_VARIANT_OPTIONS: Array<{ value: TaskGitAction; label: string }>
 
 export type RuntimeSettingsSection = "shortcuts";
 
-const SETTINGS_AGENT_ORDER: readonly RuntimeAgentId[] = ["cline", "claude", "codex", "droid"];
+const SETTINGS_AGENT_ORDER: readonly RuntimeAgentId[] = ["cline", "claude", "codex", "droid", "kiro"];
```
maxpaulus43, line 78: nit: we should prob derive this from the agent-catalog.ts or something instead

web-ui/src/components/task-start-agent-onboarding-carousel.tsx
```diff
@@ -300,7 +300,10 @@ function resolveInstallInstructions(agentId: RuntimeAgentId): string {
 		return "OpenAI's coding agent CLI with access to the latest GPT models.";
 	}
 	if (agentId === "droid") {
-		return "Factory's coding agent with access to the latest frontier models";
+		return "Factory's coding agent with access to the latest frontier models.";
+	}
+	if (agentId === "kiro") {
```
maxpaulus43, line 305: nit: this copy should live in agent-catalog.ts or some more consolidated place

src/prompts/append-system-prompt.ts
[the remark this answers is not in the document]
saoudrizwan: fixed, good callout!

src/commands/hooks.ts
saoudrizwan: everything in the hooks parsing needs to be refactored at some point... every cli hook payload is different so it makes sense to have a different parsing class for each or something like that. 

def worth spending time to get right and it be its own PR

web-ui/src/components/runtime-settings-dialog.tsx
saoudrizwan: great idea, addressed!

saoudrizwan: *actually reverted it bc im thinking we should haev a dedicated agents refactor include this as well as address the other comments you made

PR #264, opened 2026-04-07
fix: handle nullable fetchRemoteConfig in SDK boundary

johnwschoi:
Prepares for `@clinebot/core` SDK update where `fetchRemoteConfig()` returns `Promise<UserRemoteConfigResponse | null>` (backend returns `data: null` when no org has remote config).

## Changes
- Update `fetchSdkClineUserRemoteConfig` return type to `| null`
- Add `remoteConfigResponse == null` guard in `getClineKanbanAccess` before accessing response fields

## Context
- SDK PR: https://github.com/cline/sdk-wip/pull/52
- Backend PR: https://github.com/cline/core-platform/pull/2265 (merged + deployed)

The existing `try/catch` in `getClineKanbanAccess` already handles null at runtime (would throw TypeError, caught, returns `{ enabled: true }`), but this makes the null handling explicit and type-safe.

### DIFF
--- a/src/cline-sdk/cline-provider-service.ts
+++ b/src/cline-sdk/cline-provider-service.ts
@@ -452,7 +452,7 @@ export function createClineProviderService() {
 					apiBaseUrl: selectedSettings.baseUrl?.trim() || DEFAULT_CLINE_API_BASE_URL,
 					accessToken: ensureWorkosPrefix(rawAccessToken),
 				});
-				if (remoteConfigResponse == null || !remoteConfigResponse.enabled || !remoteConfigResponse.organizationId) {
+				if (!remoteConfigResponse?.enabled || !remoteConfigResponse?.organizationId) {
 					return { enabled: true };
 				}
 


src/cline-sdk/cline-provider-service.ts
```diff
@@ -452,7 +452,7 @@ export function createClineProviderService() {
 					apiBaseUrl: selectedSettings.baseUrl?.trim() || DEFAULT_CLINE_API_BASE_URL,
 					accessToken: ensureWorkosPrefix(rawAccessToken),
 				});
-				if (!remoteConfigResponse.enabled || !remoteConfigResponse.organizationId) {
+				if (remoteConfigResponse == null || !remoteConfigResponse.enabled || !remoteConfigResponse.organizationId) {
```
BarreiroT: We could just use `!remoteConfigResponse?.enabled` instead. We should also not use `==` checks.

johnwschoi: wanted to keep it explicit instead of undefined, 0 and other values. but i can update that.

BarreiroT: It shouldn't (nor it really can) be undefined, 0 or other values.

johnwschoi: updated!

johnwschoi: intent was to separate out no response vs response exists but sth else from sdk-wip; but i do think that might be too much. will note to avoid == moving forward! @BarreiroT

PR #277, opened 2026-04-09
Disable Codex startup update checks

arafatkatze:
## Summary
- disable Codex startup update checks for Kanban-launched Codex sessions by default
- preserve any explicit `check_for_update_on_startup` override already passed in agent args
- add adapter tests covering the injected default and explicit override behavior

## Why
Kanban launches Codex repeatedly for task sessions. When those launches inherit Codex's startup update check, the update prompt interrupts the session bootstrap and creates a janky startup experience inside Kanban.

The adapter already centralizes Kanban-specific Codex launch shaping, so this is the right place to force the non-interactive default while still respecting deliberate user overrides.

## Validation
- `npx vitest run test/runtime/terminal/agent-session-adapters.test.ts`
- `npm run typecheck`

### DIFF
--- a/src/terminal/agent-session-adapters.ts
+++ b/src/terminal/agent-session-adapters.ts
@@ -749,6 +749,10 @@ const codexAdapter: AgentSessionAdapter = {
 		let deferredStartupInput: string | undefined;
 		const appendedSystemPrompt = resolveHomeAgentAppendSystemPrompt(input.taskId);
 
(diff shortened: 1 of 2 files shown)


arafatkatze:
Superseded by #278, which contains the full workspace-scoped Codex host implementation.

shey-cline (approved):

PR #291, opened 2026-04-11
feat: add frontend support for two-phase WorkOS device authorization

maxpaulus43:
Add device auth types, tRPC mutations, and service methods for the new
two-phase WorkOS device authorization flow. The "cline" provider now
calls startClineDeviceAuth to get a user code and verification URL,
displays them in the UI, then polls via completeClineDeviceAuth.
Other providers (oca, openai-codex) continue using the existing
single-call OAuth flow.

Backend:
- api-contract.ts: DeviceAuthStart/Complete schemas and types
- api-validation.ts: parseClineDeviceAuthCompleteRequest
- app-router.ts: startClineDeviceAuth and completeClineDeviceAuth tRPC mutations
- runtime-api.ts: delegates to cline-provider-service
- sdk-provider-boundary.ts: boundary wrappers for SDK device auth functions
- cline-provider-service.ts: startDeviceAuth and completeDeviceAuth methods

Frontend:
- runtime-config-query.ts: startClineDeviceAuth and completeClineDeviceAuth helpers
- use-runtime-settings-cline-controller.ts: two-phase runOauthLogin, deviceAuthInfo state
- cline-setup-section.tsx: device code card UI with verification URL
- controller test: updated mocks for device auth flow

### DIFF
--- a/src/cline-sdk/cline-provider-service.ts
+++ b/src/cline-sdk/cline-provider-service.ts
@@ -8,6 +8,8 @@ import type {
 	RuntimeClineAccountOrganizationsResponse,
 	RuntimeClineAccountProfileResponse,
 	RuntimeClineAccountSwitchResponse,
+	RuntimeClineDeviceAuthCompleteResponse,
+	RuntimeClineDeviceAuthStartResponse,
 	RuntimeClineKanbanAccessResponse,
 	RuntimeClineOauthLoginResponse,
 	RuntimeClineProviderCatalogItem,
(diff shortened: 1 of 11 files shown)


saoudrizwan:
I played around with the design a bit:
Google Chrome 2026-04-13 12 16 40

saoudrizwan:
Tested this in both Docker and on a local Mac. The device-code flow makes sense for Docker/remote/headless cases where the runtime cannot open the user’s browser, but for a normal local session it would be ideal to keep the original browser OAuth flow because it is much smoother.

Is it possible to preserve the existing browser OAuth flow for clearly local/co-located users and only use device auth for remote/containerized setups? If so, I’d prefer that direction so local users do not have to enter a code unless they actually need the fallback flow.

PR #298, opened 2026-04-12
chore: refresh npm lockfiles after reinstall

finallylly:
## Summary
- refresh the root and web-ui npm lockfiles after reinstalling dependencies
- remove stale transitive entries and persist current peer dependency metadata
- keep the lockfiles aligned with the current npm resolution

## Test plan
- [x] pre-commit hook ran `tsc -p tsconfig.json --noEmit`
- [x] pre-commit hook ran `npm run test:fast`
- [x] optional: run full install/build validation if maintainers want an extra pass

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1265,29 +1265,6 @@
         "node": ">=22"
       }
     },
-    "node_modules/@emnapi/core": {
-      "version": "1.9.2",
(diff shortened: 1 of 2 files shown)


finallylly:
This PR only updates the generated lockfiles after reinstalling dependencies.

The actual package manifests were not changed, so the diff is limited to npm's current resolution metadata:
- removal of stale transitive entries that are no longer referenced
- normalization of `peer` metadata in both root and `web-ui` lockfiles

PR #300, opened 2026-04-12
fix: use accent token for inline buttons and project-row-selected

Sanitar2034:
## Changes

- Changed: `web-ui/src/components/board-card.tsx‎` : set text-accent istead of [color:inherit] for button Show more / Less in card
- Changed: `web-ui/src/styles/globals.css` : `.kb-project-row-selected` background from `#0084FF` → `var(--color-accent)`

## Why

Previously the "See more" / "Less" inline buttons inherited `color: inherit` from their parent `<p>`, blending into the secondary/tertiary text. Using the accent token makes them stand out as interactive elements while staying consistent with the design system and all theme overrides.

## Test plan
- [x] Visual check: "See more" / "Less" buttons appear in accent blue on default theme
- [x] Switch themes — buttons follow the theme accent color
- [x] Playwright tests pass confirming token-based color resolution


Sanitar2034:
Thanks for the review! Addressed in commit 95ede97:

**P1 — Cascade conflict with `[color:inherit]`**: Fixed by removing the unlayered `button.inline` rule from `globals.css` entirely. Color is now applied via Tailwind utility `text-accent` directly in `board-card.tsx`:
- Description expand/collapse buttons: `text-accent` in the static className
- Session preview expand/collapse buttons: `cn(... isTrashCard ? "text-text-tertiary" : "text-accent")` — preserves the muted treatment for trash cards

**Test assertions**: Updated `accent-color.spec.ts` to verify the accent token format (`/^#[0-9A-Fa-f]{6}$/`), non-empty background on `.kb-project-row-selected`, and that theme switching correctly changes the accent token value.

PR #317, opened 2026-04-13
Add File Tree Toggle For Diff Views

shey-cline:
Add a toggle button to hide/show the file tree panel in both the card detail diff view and the git history diff view.

**Open**
Screenshot 2026-04-13 at 3 47 00 PM

**Collapsed**
Screenshot 2026-04-13 at 3 47 07 PM

### DIFF
--- a/web-ui/src/components/card-detail-view.test.tsx
+++ b/web-ui/src/components/card-detail-view.test.tsx
@@ -72,6 +72,11 @@ vi.mock("@/stores/workspace-metadata-store", () => ({
 	useTaskWorkspaceStateVersionValue: () => 0,
 }));
 
+vi.mock("@/components/ui/tooltip", () => ({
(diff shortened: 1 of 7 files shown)


maxpaulus43:
just want to make sure this doesn't appear on a mobile view, right?

maxpaulus43 (approved):

shey-cline:
> just want to make sure this doesn't appear on a mobile view, right?

[Edited] correct

robinnewhouse:
This is great! Can we also have the diff view collapsible? :)

robinnewhouse (approved):

shey-cline:
> This is great! Can we also have the diff view collapsible? :)

Let me follow up on this.. thinking about how the ux of having toggles for full screen, file tree and diff view ties together with other panels not being collapsable

PR #326, opened 2026-04-14
Update Node to 22

BarreiroT:
Node 22 is hitting EOL. We don't need to support it much further.


abeatrix (approved):

BarreiroT:
Addressed [here](https://github.com/cline/kanban/commit/14b3d0daa5a519e1f0f22170c3a4e642d751b9f5)

PR #337, opened 2026-04-15
fix: remove live title preview from task creation, fix card click-through on title input

robinnewhouse:
## What

Two small UX fixes to the task title flow.

### 1. Remove live title preview during task creation

The "New task" dialog previously showed a Title input whose placeholder updated live as you typed the prompt. This caused the distracting effect of watching two things update simultaneously while typing.

**Now:** The title input is gone from the create dialog entirely. The title is derived from the prompt silently at save time. You can still rename any card title later via the in-place click-to-edit on the board card.

### 2. Fix click-through on the title input reopening the card

While editing a title inline on a board card, clicking inside the already-focused input to reposition the cursor was bubbling up to the card root onClick, which re-opened the card detail view.

**Root cause:** The card root onClick handler had no guard for interactive descendants, even though onMouseDownCapture already had the correct closest() guard.

**Fix:** Added the same closest("button, a, input, textarea, ...") check to onClick.

## Files changed

- `web-ui/src/App.tsx` - remove title state destructuring and prop pass-throughs
- `web-ui/src/components/task-create-dialog.tsx` - remove Title input block
- `web-ui/src/hooks/use-task-editor.ts` - remove title state; derive from prompt at save time
- `web-ui/src/components/board-card.tsx` - add closest() guard to onClick

## Screenshots

Before
Screenshot 2026-04-15 at 5 00 28 PM

After

Screenshot 2026-04-15 at 5 00 28 PM
Screenshot 2026-04-15 at 5 00 50 PM
Screenshot 2026-04-15 at 5 00 58 PM

### DIFF
--- a/src/core/task-title.ts
+++ b/src/core/task-title.ts
@@ -14,13 +14,25 @@ function truncateTaskTitle(value: string, maxChars: number): string {
 	return `${value.slice(0, maxChars).trimEnd()}…`;
 }
 
+function stripOuterUserInputTag(value: string): string {
+	return value
+		.replace(/^<user_input(?:\s[^>]*)?>/u, "")
+		.replace(/<\/user_input>$/u, "")
+		.trim();
+}
+
 export function deriveTaskTitleFromPrompt(prompt: string, maxChars = DEFAULT_TASK_TITLE_MAX_CHARS): string {
 	const firstNonEmptyLine =
 		prompt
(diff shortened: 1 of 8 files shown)


robinnewhouse:
@greptileai

shey-cline (approved):
Cool tasks only

PR #345, opened 2026-04-16
fix(sdk): prevent duplication of tool call arguments in assistant message history (CLINE-1825)

marklapasaDS:
**Description**
Fixed a bug where tool call arguments were being duplicated in the `raw_arguments` field, leading to JSON parse errors.

**The Problem**
In the Cline SDK event stream, tool arguments are sent as JSON chunks before the structured `content_start` event arrives. The `cline-event-adapter.ts` was treating these JSON chunks as regular assistant text because the `isLikelySerializedAgentEventChunk` filter only excluded JSON objects that contained a `"type"` field. Since tool arguments are plain JSON objects without a `"type"` field, they were appended to the assistant's message history. When the SDK later assembled the final tool call, it concatenated these accumulated chunks with the structured input, resulting in duplicated JSON objects (e.g., `{"files":[...]}{"files":[...]}`) which is invalid JSON.

**The Solution**
Updated `isLikelySerializedAgentEventChunk` in `src/cline-sdk/cline-event-adapter.ts` to filter out any valid JSON object sent as a chunk in the agent stream. This ensures that tool arguments and other serialized events are not accidentally treated as chat text.

**Changes**
- Modified `src/cline-sdk/cline-event-adapter.ts`: Updated `isLikelySerializedAgentEventChunk` to return `true` for any valid JSON object.

**Verification Results**
- **Reproduction Test:** Created `test/runtime/cline-sdk/repro-duplication.test.ts` which specifically simulates the sequence of chunk events followed by a tool event. The test now passes, confirming that tool argument chunks are no longer appended to the assistant's message history.
- **Regression Testing:** Ran the entire `test/runtime/cline-sdk/` test suite (94 tests), including `cline-event-adapter.test.ts` and `cline-task-session-service.test.ts`. All tests passed, ensuring no regressions were introduced.

**Related Ticket**
Fixes `CLINE-1825`


marklapasaDS:
No longer reproducible in the latest [v0.1.64](https://github.com/cline/kanban/releases/tag/v0.1.64)

PR #352, opened 2026-04-17
fix(pre-commit): use `git add -f` so stale info/exclude can't block commits

### LINKED ISSUE #351: Stale /packages entry in .git/info/exclude blocks git add and trips husky pre-commit hook
## Summary

When working on files under `packages/desktop/`, `git add packages/desktop/src/...` can fail with:

```
The following paths are ignored by one of your .gitignore files:
packages
hint: Use -f if you really want to add them.
```

…even though the committed `.gitignore` only ignores `packages/desktop/out` and `packages/desktop/*.tgz`. This also breaks the husky pre-commit hook's final `git add "$file"` loop, causing the whole commit to fail with `husky - pre-commit script failed (code 1)`.

## Root cause

The block Kanban maintains in `.git/info/exclude` (via `syncManagedIgnoredPathExcludes` in `src/workspace/task-worktree.ts`) can end up with a stale `/packages` entry:

```
# kanban-managed-symlinked-ignored-paths:start
# Keep symlinked ignored paths ignored inside Kanban task worktrees.
/.env
/dist
/node_modules
/packages        ← stale; shadows all tracked files under packages/
/.husky/_
/web-ui/dist
/web-ui/node_modules
# kanban-managed-symlinked-ignored-paths:end
```

That entry was written by an older Kanban run — when `packages/` didn't yet contain any tracked content. Running `git ls-files --others --ignored --exclude-per-directory=.gitignore --directory` today (the input to `listIgnoredPaths`) returns granular sub-paths like `packages/desktop/dist/`, `packages/desktop/node_modules/`, `packages/desktop/out/` — but never `packages/` on its own — so a fresh regeneration would not produce the broken line.

However `syncManagedIgnoredPathExcludes` is only invoked during task-worktree setup, and `.git/info/exclude` is shared across `git worktree` worktrees via `git-common-dir`. Without a triggering worktree operation the stale line lingers indefinitely and silently breaks basic `git add` in any worktree on the same clone.

## Repro (approximate)

1. Clone this repo at a commit where `packages/` has no tracked content (or pre-dates the `packages/desktop` package).
2. Let Kanban run its worktree setup once so `syncManagedIgnoredPathExcludes` writes `/packages` to `.git/info/exclude`.
3. Pull / switch to a branch that has tracked files under `packages/`.
4. `git add packages/desktop/src/any-tracked-file.ts` → fails with the "paths ignored" error. Commit via husky fails with exit 1.

## Workaround

Delete the stale line:

```bash
sed -i.bak '/^\/packages$/d' "$(git rev-parse --git-common-dir)/info/exclude"
```

### 1. Harden `syncManagedIgnoredPathExcludes` (`src/workspace/task-worktree.ts`)

Filter out any path that has tracked content underneath before writing it to `.git/info/exclude`. Even if `git ls-files` or `getUniquePaths` somehow produces an ancestor path of a tracked file, we should never write an exclude line that would shadow tracked content.

Sketch:

```ts
async function filterPathsShadowingTrackedContent(
  repoPath: string,
  relativePaths: string[],
): Promise<string[]> {
  const kept: string[] = [];
  for (const p of relativePaths) {
    const hasTracked = await getGitStdout(
      ["ls-files", "-z", "--", p],
      repoPath,
    );
    if (!hasTracked) kept.push(p);
  }
  return kept;
}
```

…and call it inside `syncManagedIgnoredPathExcludes` before serializing the block.

### 2. Harden the husky pre-commit hook (`.husky/pre-commit`)

The re-add loop currently does:

```sh
for file in $STAGED_FILES; do
    if [ -f "$file" ]; then
        git add "$file"
    fi
done
```

Switch to `git add -f "$file"` — these files were just staged, so forcing is safe, and it survives any future stale-exclude situation (from Kanban or any other tool).

## Impact

Any developer working on `packages/desktop/` in a clone that picked up the stale entry can't commit normally — they have to either know the `info/exclude` workaround or use `--no-verify`. Low-visibility footgun that only shows up for folks touching the Electron package.

## Related

Surfaced while reviewing the Desktop PR stack (`pr/desktop-2-webui-compat`, `pr/desktop-3-scaffolding`, `pr/desktop-4-electron-main`). The fix for the PR stack itself used `--no-verify` for one commit (`8f53c463` on `pr/desktop-4-electron-main`); the underlying stale entry was removed manually from the local clone.

johnwschoi:
Closes #351.

## Change

```diff
 for file in $STAGED_FILES; do
 	if [ -f "$file" ]; then
-		git add "$file"
+		git add -f "$file"
 	fi
 done
```

One line in `.husky/pre-commit`.

## Why this is a real bug

`git add <path>` refuses to add a file whose parent directory is matched by `.git/info/exclude` (or `.gitignore`), **even if the file is already tracked**. Reproduced in a scratch repo:

```console
$ # repo has packages/foo.ts tracked, /packages in .git/info/exclude
$ git add packages/foo.ts
The following paths are ignored by one of your .gitignore files:
packages
hint: Use -f if you really want to add them.
# exit 128
```

That's exactly the hook's failure mode. The re-stage loop runs *after* `biome check --staged` has applied fixes in-place and is how the hook pulls those fixes back into the index. If the developer has a stale Kanban-managed `/<dir>` line in `.git/info/exclude` (written by an older task-worktree run before that directory gained tracked content), every `git add "$file"` in the loop hits `fatal: ignored`, husky exits 1, and the developer sees `pre-commit script failed (code 1)` with no useful signal — even though biome, typecheck, and tests all passed.

## Why `-f` is safe here

- The loop only re-adds paths that were already `git add`-ed at the start of the hook, i.e. files the developer explicitly chose to stage.
- `-f` only overrides ignore rules; it does not touch permissions, symlinks, large-file/LFS hooks, or any other git policy.
- It matches user intent: a commit the developer authored shouldn't be silently vetoed by a stale client-side ignore rule.

## Scope

Just the hook. No source or test changes — `syncManagedIgnoredPathExcludes` in `src/workspace/task-worktree.ts` already regenerates the managed exclude block correctly from fresh `git ls-files --others --ignored` output (which, by definition, can never include paths that shadow tracked content), so adding "defensive" filtering there would be dead code guarding against impossible input.

## Verification

- `npm run typecheck` ✅
- `npm run test:precommit` ✅ 504 / 504
- `npx biome check` ✅
- **End-to-end**: re-injected `/packages` into a local `.git/info/exclude`, then `git commit` with the full husky hook enabled — hook passed, commit landed. Without the `-f`, the same commit fails as in the scratch-repo repro above.

### DIFF
--- a/.husky/pre-commit
+++ b/.husky/pre-commit
@@ -16,9 +16,11 @@ if [ $? -ne 0 ]; then
 	exit 1
 fi
 
+# Use -f so an over-broad .git/info/exclude entry can't block the re-add;
+# these files were already staged, so forcing is safe.
 for file in $STAGED_FILES; do
 	if [ -f "$file" ]; then
-		git add "$file"
+		git add -f "$file"
 	fi
 done
 


johnwschoi:
@greptileai

johnwschoi:
Closing as not worth shipping.

conclusion is: the bug is transitional and essentially me-only in practice.

**Why:**

- `packages/` does not exist on `origin/main`. It only appears on the in-flight desktop PR branches.
- For `/packages` to land in anyone's `.git/info/exclude` managed block, `git ls-files --others --ignored --directory` must emit bare `packages`, which requires (a) `packages/` to exist on disk, and (b) nothing tracked inside it.
- On main-based workflows: (a) fails — `packages/` isn't there — so no dev cloning/pulling main can end up with `/packages` in their managed block.
- The original hit was me running task-worktree setup during local desktop-stack iteration while `packages/desktop/` existed on disk with only build artifacts (`node_modules/`, `.tgz`, `dist/`, `out/`) before any source was committed. `syncManagedIgnoredPathExcludes` correctly wrote `/packages` at that moment. Once `packages/desktop/src/...` became tracked content, the entry went stale.
- The managed block self-heals the next time `syncManagedIgnoredPathExcludes` runs. Mine already did earlier today.

`git add -f` in the re-stage loop is still the semantically correct verb (re-staging already-staged files shouldn't be ignore-gated), but the cost of shipping this defense-in-depth isn't justified by the impact. Closing.

PR #359, opened 2026-04-17
fix(web-ui): keep Add Project button mounted during project switch

### LINKED ISSUE #343: Visual flicker of "Add Project" button when switching between projects
### What happened?

Describe the bug When switching between projects under the "Projects" section, the "Add Project" button briefly disappears and then re-renders along with the swimlane panel refresh. This causes a noticeable visual flicker.

### Steps to reproduce

1. Go to the 'Projects' view.
2. Click to switch from one project to another.
3. Observe the "Add Project" button area during the transition.

### Expected behavior

The "Add Project" button should remain static and unaffected during project switches. The user should not perceive any visual change (like disappearance and reappearance) of this button.

### Kanban version

_No response_

### Environment

macOS 14.5

Aiudadadadf:
## Summary

Fixes #343.

The **Add Project** button briefly unmounted and re-mounted every time the user switched projects, producing a visible flicker alongside the swimlane refresh.

## Root cause

`ProjectNavigationPanel` gates the Add Project button on `!isLoadingProjects`. That prop is driven by `isProjectListLoading = !hasReceivedSnapshot && !streamError` from `use-project-ui-state.ts`. When the user switches projects, the `requested_workspace_changed` action in `use-runtime-state-stream.ts` resets `hasReceivedSnapshot` to `false`:

```ts
// web-ui/src/runtime/use-runtime-state-stream.ts
if (action.type === "requested_workspace_changed") {
    return {
        ...state,
        // ...
        hasReceivedSnapshot: false,
    };
}
```

That flip is brief — the next snapshot arrives almost immediately — but long enough to unmount the button and remount it, causing the flicker.

## Fix

The existing gate `!isLoadingProjects` conflated "initial load (no projects yet)" with "project switch (projects already populated)". The skeleton placeholders above already cover the initial-load case (`sortedProjects.length === 0 && isLoadingProjects`), so the button only needs to be hidden in that specific state. The gate now mirrors the skeleton condition:

```tsx
{sortedProjects.length === 0 && isLoadingProjects ? null : (
    <button ...>Add Project</button>
)}
```

During a project switch `sortedProjects.length > 0` (the project list is preserved across the switch), so the button stays mounted — no flicker. During the genuine initial load it still hides behind the skeletons as before.

## Tests

Two regression tests added to `project-navigation-panel.test.tsx`:

- `keeps the Add Project button rendered while the project list reloads after a switch` — asserts the button is present when `isLoadingProjects=true` and projects are populated.
- `hides the Add Project button on initial load while projects are still being fetched` — asserts the original behavior still holds when there are no projects yet.

Full web-ui test suite: **491 passed**.

### DIFF
--- a/web-ui/src/components/project-navigation-panel.test.tsx
+++ b/web-ui/src/components/project-navigation-panel.test.tsx
@@ -211,6 +211,16 @@ describe("ProjectNavigationPanel width persistence", () => {
 		expect(container.textContent).not.toContain("Report issue");
 	});
 
+	it("keeps the Add Project button rendered while the project list reloads after a switch", () => {
+		renderPanel({ isLoadingProjects: true });
+		expect(() => getButtonByText(container, "Add Project")).not.toThrow();
+	});
+
+	it("hides the Add Project button on initial load while projects are still being fetched", () => {
+		renderPanel({ projects: [], currentProjectId: null, isLoadingProjects: true });
+		expect(Array.from(container.querySelectorAll("button")).some((b) => b.textContent === "Add Project")).toBe(false);
+	});
+
 	it("persists terminal tips dismissal", () => {
 		renderPanel({
 			activeSection: "agent",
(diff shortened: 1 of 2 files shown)


TyroneNel:
Re-implemented this on current main as https://github.com/cline/kanban/pull/598 so the flicker fix can land without updating this stale branch.

PR #355, opened 2026-04-17
[1/5] feat(desktop): bootstrap @kanban/desktop workspace + bundle CLI runtime deps

johnwschoi:
## Summary

First of a 5-PR stack that adds Electron desktop support. Pure bootstrap — no Electron code yet, just the workspace scaffolding and a build-infrastructure prerequisite.

**Follow-up PRs** (branches pushed, will open one at a time):
- `pr/desktop-4-runtime-child` — runtime child process manager
- `pr/desktop-5-preflight` — desktop preflight checks
- `pr/desktop-6-electron-main` — Electron main process
- `pr/desktop-7-packaging` — electron-builder packaging + CLI shims

### 1. [`7e719213`](https://github.com/cline/kanban/pull/355/commits/7e719213) — `build: bundle all non-native runtime deps into dist/cli.js`

The previous `scripts/build.mjs` externalized 11 packages at bundle time (`zod`, `commander`, `@trpc/client`, `@trpc/server`, `@sentry/node`, `@modelcontextprotocol/sdk`, `ws`, `open`, `proper-lockfile`, `tree-kill`, `node-pty`), so `dist/cli.js` contained literal `import "zod"` / `require("ws")` that Node resolved via an ancestor `node_modules/` at runtime.

That worked for the two existing contexts:
1. **`npm i -g kanban`** — npm installs prod deps at `lib/node_modules/kanban/node_modules/`; resolver walks up from `dist/cli.js` and finds them.
2. **Monorepo dev** — root `node_modules/` satisfies resolution.

But it fails for the packaged Electron app: `Resources/cli/cli.js` has no enclosing `node_modules/` → `ERR_MODULE_NOT_FOUND`.

**Fix**: drop the externals list down to just `node-pty` (the one genuinely non-bundlable package — native addon with a compiled binding + spawn-helper binary on disk). esbuild inlines the other 10 directly into the bundle.

**Result**: `dist/cli.js` is a single self-contained 16 MB ESM bundle with one runtime-resolved external. In the packaging PR later in this stack, `node-pty` gets rebuilt for Electron's ABI by `electron-builder install-app-deps` and the CLI's ESM resolution walks up from its location in `app.asar.unpacked/cli/` to find it in `app.asar.unpacked/node_modules/`.

[...]

PR #370, opened 2026-04-20
[2/5] feat(desktop): runtime child process manager

johnwschoi:
Part 2 of the desktop split. Stacked on #355.

Introduces `RuntimeChildManager` — spawns the Kanban CLI as a subprocess and manages its lifecycle:

- Spawns the CLI out-of-process (no in-process runtime imports)
- HTTP readiness polling against `/` (default 200ms, 30s total timeout)
- Typed lifecycle events: `ready`, `error`, `crashed`
- Clean contract: failures before `ready` reject `start()`; failures after `ready` flow through events. One failure never produces both
- Graceful shutdown via SIGTERM, force-kill fallback after 5s (`taskkill /T /F` on Windows)
- Rolling 8KB stderr tail surfaced to the `crashed` listener for diagnostics

Env/PATH policy lives in its own file (`runtime-child-env.ts`):

- Strict allowlist: exact keys (PATH, HOME, LANG, SHELL, XDG_*, Windows %APPDATA% family, etc.) + prefix allowlist (KANBAN_, ANTHROPIC_, OPENAI_, etc.)
- GUI-launch PATH enrichment: Homebrew on macOS, `/snap/bin` on Linux, npm-global/Node/Git on Windows

Tests: 40 unit tests covering spawn, ready-path, crash scenarios, shutdown races, env filtering, and PATH assembly.


johnwschoi:
@greptileai

packages/desktop/src/runtime-child-env.ts
```diff
@@ -0,0 +1,147 @@
...
+	"ProgramFiles(x86)",
+	"ProgramData",
+	"SYSTEMDRIVE",
+	"XDG_RUNTIME_DIR",
+	"XDG_CONFIG_HOME",
+	"XDG_DATA_HOME",
+]);
+
+/** Prefixes forwarded verbatim — provider API keys and KANBAN_* config. */
+export const ALLOWED_ENV_PREFIXES: readonly string[] = [
```
maxpaulus43, line 49: this seems brittle, is there a better way to support more generic keys?

packages/desktop/src/runtime-child.ts
```diff
@@ -0,0 +1,327 @@
...
+ *
+ *   - Windows: `taskkill /T /F` walks the tree.
+ *   - POSIX:   signals the whole process group (`-pid`). Requires the child
+ *              to have been spawned with `detached: true` so it leads its own
+ *              process group; otherwise `-pid` would target *our* group.
+ */
+function treeKill(pid: number, signal: NodeJS.Signals = "SIGTERM"): void {
+	if (process.platform === "win32") {
+		try {
+			execSync(`taskkill /T /F /PID ${pid}`, { stdio: "ignore" });
```
maxpaulus43, line 82: so we need elevated priveleges to run this command?

packages/desktop/src/runtime-child.ts
```diff
@@ -0,0 +1,327 @@
...
+				req.destroy();
+				setTimeout(check, pollIntervalMs);
+			});
+		};
+
+		check();
+	});
+}
+
+export class RuntimeChildManager extends EventEmitter<RuntimeChildManagerEvents> {
```
maxpaulus43, line 143: this PR doesn't call this Class at all yet, right?

packages/desktop/src/runtime-child.ts
```diff
@@ -0,0 +1,327 @@
...
+		this.disposed = true;
+		await this.shutdown();
+		this.removeAllListeners();
+	}
+
+	get running(): boolean {
+		return this.child !== null;
+	}
+
+	private async spawnChild(config: RuntimeChildConfig): Promise<string> {
```
maxpaulus43, line 211: feels like we do very similar stuff in cline with background terminals. anything we could/should re-use there?

packages/desktop/src/runtime-child.ts
johnwschoi: Yup — #372 (Electron main) wires it up; this PR just introduces the class as part of the stacked series.

packages/desktop/src/runtime-child.ts
johnwschoi: Runtime lifecycle like HTTP health checks, crash recovery, ready state, is  different from user shell PTY sessions unfortunately

packages/desktop/src/runtime-child.ts
johnwschoi: we dont need it since /F and /T are target process owned by you so elevation is only required if kill process owned by other users

packages/desktop/src/runtime-child-env.ts
johnwschoi: ah i was trying to be more intentially strict and deny by default so that we only allow knwn-safe keys plus ai provider prefixes. i am open to suggestions here though!

maxpaulus43: if we want to support new providers, we will need to update this? What could happen if we don't deny by default?

maxpaulus43 (approved):

packages/desktop/src/runtime-child-env.ts
johnwschoi: it is actually a good suggestion, hmmm so if we dont do deny by default, we can expose credentials in env vars  (github token, npm token, stripe key etc) or forward non ai cloud services like vercel or sth which we might not want. But that does come at a cost of needing to update new providers, as you mentioned.I think we can revisit this, its definitely valid

PR #371, opened 2026-04-20
refactor(desktop): simplify env handling and remove over-engineered checks

johnwschoi:
Simplifies desktop package by removing over-engineered validation and switching to allow-all env forwarding.

**Changes:**
- Forward all parent env vars by default (matches VS Code's model)
- Merge parent NODE_OPTIONS instead of overwriting (preserves flags like --require)
- Handle both --max-old-space-size and --max_old_space_size variants
- Remove entire preflight validation system (doesn't add value)
- Remove unused node-pty dependency

**Rationale:**
- Env filtering: Agent commands are user-approved, so allowlisting provides no real security boundary
- NODE_OPTIONS: Preserves parent flags in Electron deployments
- Preflight: BrowserWindow/spawn fail immediately with clear errors anyway; preflight just moves the error 10ms earlier
- node-pty: Unused after removing preflight check; removes 455 packages

**Lines removed:** ~854 lines of defensive infrastructure
**Tests:** All pass (40/40)


johnwschoi:
@greptileai

packages/desktop/src/desktop-preflight.ts
```diff
@@ -0,0 +1,128 @@
...
+	/** Soft failures — the app will boot but some features may be degraded. */
+	warnings: DesktopPreflightWarning[];
+	resources: {
+		preloadExists: boolean;
+		cliShimExists: boolean;
+		nodePtyLoadable: boolean | null;
+	};
+}
+
+export function runDesktopPreflight(
```
maxpaulus43: why would we need this? how could it be that we vend out an app that doesn't have the libraries bundled with it?

johnwschoi: gave it some thought, and decided to remove it! preflight check would mostly be helpful during dev/testing but not necessarily for prod! great suggestion

packages/desktop/src/runtime-child-env.ts
```diff
@@ -15,59 +15,6 @@
 
 import path, { join } from "node:path";
 
-/** Exact env var names forwarded verbatim from the parent process. */
-export const ALLOWED_ENV_KEYS: ReadonlySet<string> = new Set([
```
johnwschoi, line 19: @maxpaulus43  gave it second thought and removed it. this is same behavior as vscode

maxpaulus43 (approved):

PR #383, opened 2026-04-22
Cline SDK 0.0.36

abeatrix:
Drop the `@clinebot/agents` and `@clinebot/llms` packages as direct dependencies, retaining only `@clinebot/core` (bumped to ^0.0.36) and `@clinebot/shared`. Update the Biome import restriction rule to only guard `@clinebot/core` imports, and revise AGENTS.md to reflect the simplified dependency surface.


BarreiroT (approved):
Mhmm, not too happy about the type casting, but LGTM to unblock

PR #406, opened 2026-04-24
Add LiteLLM model refresh action

arafatkatze:
## Motivation
LiteLLM and other Base URL-backed providers get their model list from saved SDK provider settings. Before this change, editing the Base URL in Kanban settings and immediately refreshing models still used the previous saved settings, so the user had to save, reopen settings, and only then see aliases from the new LiteLLM URL.

The refresh action now saves the current draft provider settings first, then asks the runtime for models. That keeps the SDK-backed provider store, Kanban UI state, and keyless LiteLLM `/v1/models` discovery pointed at the same URL during the refresh.

## Summary
- add a small refresh action for Base URL-backed Cline providers so settings are saved before fetching provider models
- refresh provider models after saving the current LiteLLM/custom provider base URL and API key draft values
- keep the settings dialog on Cline after refreshing models instead of resetting to the saved default agent
- add a compact LiteLLM keyless Base URL discovery fallback so local proxies exposing `/v1/models` show their model aliases even when no API key is configured
- bound the LiteLLM model-list probe to 2.5s, preserve configured request headers/global fetch behavior, and parse `/model/info` aliases from `model_name`
- guard provider-model loads with a latest-request id so an older initial load cannot overwrite models returned by a newer refresh

## Verification
- verified `http://127.0.0.1:4000/v1/models` returns `litellm-test-alpha`, `litellm-test-beta`, and `litellm-test-gamma`
- verified Kanban runtime returns `gpt-5.4` plus those three LiteLLM aliases for `getClineProviderModels({ providerId: "litellm" })` with `apiKeyConfigured: false`
- `npx @biomejs/biome check src/cline-sdk/cline-provider-service.ts test/runtime/cline-sdk/account-balance.test.ts test/runtime/trpc/runtime-api.test.ts web-ui/src/components/shared/cline-setup-section.tsx web-ui/src/hooks/use-runtime-settings-cline-controller.ts web-ui/src/hooks/use-runtime-settings-cline-controller.test.tsx`
- `npx vitest run test/runtime/cline-sdk/account-balance.test.ts`
- `npm --prefix web-ui run test -- src/hooks/use-runtime-settings-cline-controller.test.tsx`
- `npm run typecheck`
- `npm run test:fast`
- `npm run web:build`
- `npm --prefix packages/desktop run typecheck`
- `npm --prefix packages/desktop test`
- pre-commit: Biome, root typecheck, `npm run test:fast`

### DIFF
--- a/src/cline-sdk/cline-provider-service.ts
+++ b/src/cline-sdk/cline-provider-service.ts
@@ -21,6 +21,7 @@ import type {
 	RuntimeClineReasoningEffort,
 } from "../core/api-contract";
 import { openInBrowser } from "../server/browser";
+import { createKanbanClineLogger } from "./cline-runtime-logger";
 import {
 	addSdkCustomProvider,
 	completeClineDeviceAuth as completeSdkDeviceAuth,
@@ -58,8 +59,16 @@ const MANAGED_PROVIDER_ENV_KEYS: Record<ManagedClineOauthProviderId, readonly st
 const CLINE_REMOTE_CONFIG_SCHEMA = z.object({
 	kanbanEnabled: z.boolean().optional(),
 });
+const LITELLM_MODELS_RESPONSE_SCHEMA = z.object({
+	data: z.array(z.object({ id: z.string().optional(), model_name: z.string().optional() }).passthrough()).optional(),
+});
+const LITELLM_MODEL_LIST_PATHNAMES = ["/models", "/model/info"] as const;
+const LITELLM_MODEL_LIST_TIMEOUT_MS = 2_500;
+const LOGGER = createKanbanClineLogger({ component: "cline-provider-service" });
 
 type ClineRemoteConfig = z.infer<typeof CLINE_REMOTE_CONFIG_SCHEMA>;
+type LiteLlmModelListPathname = (typeof LITELLM_MODEL_LIST_PATHNAMES)[number];
+type LiteLlmModelListItem = NonNullable<z.infer<typeof LITELLM_MODELS_RESPONSE_SCHEMA>["data"]>[number];
 type SdkReasoningEffort = NonNullable<NonNullable<SdkProviderSettings["reasoning"]>["effort"]>;
 
 export interface ResolvedClineLaunchConfig {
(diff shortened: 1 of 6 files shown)


src/cline-sdk/cline-provider-service.ts
```diff
@@ -227,6 +230,37 @@ function toRuntimeProviderModel(model: RuntimeClineProviderModel): RuntimeClineP
...
+	}
+
+	const apiKey = resolveVisibleApiKey(settings);
+	const headers: Record<string, string> = {};
+	if (apiKey) {
+		headers.Authorization = `Bearer ${apiKey}`;
+	}
+	for (const pathname of ["/models", "/model/info"] as const) {
+		try {
+			const response = await fetch(`${baseUrl.replace(/\/+$/, "")}${pathname}`, { method: "GET", headers });
```
dominiccooney: Is this global fetch? How does fetch patching work for SDK consumers who need to support proxies?

src/cline-sdk/cline-provider-service.ts
```diff
@@ -227,6 +230,37 @@ function toRuntimeProviderModel(model: RuntimeClineProviderModel): RuntimeClineP
...
+			const parsed = response.ok
+				? LITELLM_MODELS_RESPONSE_SCHEMA.safeParse((await response.json()) as unknown)
+				: null;
+			const modelIds =
+				parsed?.success === true
+					? parsed.data.data?.map((item) => (item.id ?? item.model_name ?? "").trim()).filter(Boolean)
+					: [];
+			if (modelIds && modelIds.length > 0) {
+				return [...new Set(modelIds)].map((id) => ({ id, name: id }));
```
dominiccooney, line 304: Is it right to, given items with id and model_name, throw away the model name when returning id and name?

src/cline-sdk/cline-provider-service.ts
```diff
@@ -227,6 +230,37 @@ function toRuntimeProviderModel(model: RuntimeClineProviderModel): RuntimeClineP
...
+			}
+		} catch {
+			// Fall back to the next LiteLLM-compatible model listing endpoint.
```
dominiccooney: Is there any way we can tighten up what's swallowed here? Because code like this gives the `try {` block an invisibility cloak to our error monitoring, etc.

src/cline-sdk/cline-provider-service.ts
```diff
@@ -786,13 +820,21 @@ export function createClineProviderService() {
...
 			const normalizedProviderId = providerId.trim().toLowerCase();
-			const providerModels =
+			let providerModels =
 				normalizedProviderId.length > 0
 					? await listSdkProviderModels(normalizedProviderId)
 							.then((sdkModels) => sdkModels.map((model) => toRuntimeProviderModel(model)))
 							.then((sdkModels) => sdkModels.sort((left, right) => left.name.localeCompare(right.name)))
 							.catch(() => [])
 					: [];
+			if (normalizedProviderId === "litellm") {
```
dominiccooney, line 882: You know this before you do listSdkProviderModels, so you may as well start the fetchLiteLlmBaseUrlModels straight away.

dominiccooney (dismissed):
Some feedback inline.

Could you put a description in the PR which explains the motivation for the change? The description that is there seems to be mechanical about how it is meant to work.

Why do we want to save settings first before fetching, for example?

What do you think of greptile's comment about slow fetches? This code is pretty typical for what we write, but it does seem very optimistic about operations succeeding quickly.

arafatkatze:
@dominiccooney thanks, I updated the PR description with a Motivation section.

The reason for save-before-fetch is that the runtime model lookup reads from the SDK-persisted provider settings, not the unsaved settings form draft. Without saving first, the refresh still probes the previously saved Base URL, which is why LiteLLM aliases only appeared after saving and reopening settings.

On slow fetches: agreed. I added a 2.5s `AbortSignal.timeout(...)` around the LiteLLM model-list probe, and failures now go through the Kanban Cline logger instead of an empty catch. The fallback still tries `/models` then `/model/info`, but a dead or blackholed Base URL is now bounded rather than waiting for OS-level network timeout.

src/cline-sdk/cline-provider-service.ts
arafatkatze: Updated this to call `globalThis.fetch` explicitly and to preserve configured provider `headers` plus the resolved `Authorization` header. I checked the SDK-facing provider settings/config surface here; this layer does not expose a per-provider fetch override, so setups that patch global fetch for proxy behavior remain the path this fallback follows. The runtime test now stubs global fetch and verifies configured headers pass through.

src/cline-sdk/cline-provider-service.ts
arafatkatze: Changed this to be endpoint-specific. `/models` uses `id`; `/model/info` now prefers `model_name` and only falls back to `id` if `model_name` is absent. The returned `RuntimeClineProviderModel` still uses `{ id, name: id }` because this fallback is feeding the selectable model alias, but we no longer discard `model_name` when LiteLLM provides it. Added a test with `id: "internal-id"` and `model_name: "litellm-test-alias"` to cover this.

src/cline-sdk/cline-provider-service.ts
arafatkatze: Tightened this up. The empty catch is gone: fetch exceptions are caught as `error` and logged through `createKanbanClineLogger` with the URL and `errorMessage`. Non-OK HTTP responses and unexpected response shapes also emit warning metadata before falling through to the next LiteLLM endpoint.

src/cline-sdk/cline-provider-service.ts
arafatkatze: I left this serial for now after adding the timeout. Starting the LiteLLM probe earlier is straightforward, but it also means every LiteLLM model load immediately schedules an external Base URL request even when SDK/local models are already available. I kept this PR to the smaller refresh/fallback behavior change and bounded the bad-URL case with a 2.5s timeout. Happy to switch to parallel probing if you prefer that latency tradeoff.

robinnewhouse:
Quick feedback from Cline's review:

The model refresh path and the initial model-load effect can race each other. If settings opens and starts loading models for the old saved LiteLLM URL, then the user edits the Base URL and clicks refresh, the refresh can save/fetch the new models but still be overwritten by the older in-flight request when it resolves later. Could we add a shared request id / latest-request guard, or route both paths through one loader so only the newest model request updates `providerModels`?

arafatkatze:
@robinnewhouse good catch. I pushed a fix for this in `21f873a5`.

The Cline settings hook now routes model fetches through a shared latest-request guard. Refresh invalidates any older in-flight model load before saving/fetching, and only the newest request is allowed to update `providerModels` or clear the loading state. I also reused that guarded loader for the add/update custom provider paths so they follow the same state write rule.

Added a regression test that starts the initial LiteLLM model load, refreshes models from the edited Base URL, then resolves the original initial request afterward and verifies the stale models do not overwrite the refreshed list.

robinnewhouse (approved):

PR #408, opened 2026-04-24
feat(update): show "new version available" modal with one-click update [ENG-1686]

robinnewhouse:
Adds a modal that pops up in the web UI when a newer Kanban version is available on npm, with a one-click "Update Now" button that runs the right install command for the user's package manager. Closes ENG-1686.

## Why

Kanban is meant to run all the time, so users naturally fall behind on versions. There was already a backend startup auto-update check (`runAutoUpdateCheck`) plus a `kanban update` CLI command (`runOnDemandUpdate`), but neither surfaced anything in the UI — a user with the browser tab open all day had no signal that a new version had shipped, and the `--update` flag is invisible to non-CLI users.

This adds the missing UI piece: when the existing startup check finds a newer version, the next page render shows a centered modal that mirrors the linked design (icon, title, current/latest version, copy-able install command, Later / Update Now). Update Now goes through the existing `runOnDemandUpdate`, so behavior matches `kanban update` exactly — same package-manager detection, same fallbacks.

## Change

**Backend**

- `src/update/update.ts`: extend `runAutoUpdateCheck` to record a `PendingUpdateNotification` (current/latest version, update timing, user-facing install command) when a newer version is detected. Add `buildUserFacingInstallCommand` so the modal shows `pnpm add -g …`, `npx kanban`, etc. — not the internal cache-refresh script.
- `src/core/api-contract.ts`: add `runtimeUpdateStatusResponseSchema` and `runtimeRunUpdateResponseSchema`.
- `src/trpc/runtime-api.ts` + `src/trpc/app-router.ts`: expose `runtime.getUpdateStatus` (query) and `runtime.runUpdateNow` (mutation).
- `src/server/runtime-server.ts` + `src/cli.ts`: thread `getUpdateStatus` and `runUpdateNow` as deps. `runUpdateNow` delegates to the existing `runOnDemandUpdate` so users get the same install behavior as `kanban update`.

**Frontend**

- `web-ui/src/runtime/runtime-config-query.ts`: `fetchRuntimeUpdateStatus` and `runRuntimeUpdateNow` tRPC client helpers.
- `web-ui/src/hooks/use-update-notification.ts`: polls the runtime once on mount, retries after 10s (the server populates the pending notification asynchronously after startup), then falls back to an hourly poll. Returns `{ availableUpdate, dismiss }` and is session-scoped — dismissing won't resurface the same prompt until the next page load.
- `web-ui/src/components/update-available-dialog.tsx`: centered modal with the install command in a copy-able code block. Phases: idle / running (spinner on Update Now, Later disabled) / success (green check + "Restart Kanban to use the new version" + Close) / error (inline red alert, retryable).
- `web-ui/src/App.tsx`: wire the hook and render the modal alongside the other top-level dialogs.

**Tests**

- `test/runtime/update/auto-update.test.ts`: extend the `getPendingUpdateNotification` suite to assert `installCommand` for both startup (`npm install -g …`) and shutdown (`npx kanban`) timing paths.

## How to test

1. Temporarily seed a pending notification at the top of `src/update/update.ts`:
   ```ts
   let pendingUpdateNotification: PendingUpdateNotification | null = {
       currentVersion: "0.1.64",
       latestVersion: "0.1.99",
       updateTiming: "startup",
       installCommand: "npm install -g kanban@latest",
   };
   ```
2. `npm run dev:full`, open the Vite dev URL.
3. Modal appears within ~1s. Click the copy icon — command copies. Click **Update Now** — spinner shows, `npm install -g kanban@latest` runs in the dev:full terminal, modal flips to a green success state. Click **Close**.
4. Revert the seeded value (it is `null` on this branch).

In production, the modal only fires when the existing startup auto-update check actually finds a newer version on npm, so end users see it organically.

## Notes

- Update behavior is unchanged from `kanban update`. We didn't add a parallel install path; the modal just calls into `runOnDemandUpdate`.
- The hook only polls hourly after the initial check + 10s retry. The runtime's `runAutoUpdateCheck` only runs on startup, so polling more often wouldn't produce different results.
- `Later` dismisses the modal for the session. The next `kanban` launch will show it again if the underlying version still mismatches.

## Screenshots
Screenshot 2026-04-24 at 3 26 36 PM
Screenshot 2026-04-24 at 3 26 37 PM


robinnewhouse:
Thanks for the review. Both points addressed in `b95daeff`:

**P1 — `clearPendingUpdateNotification()` after success.** `src/cli.ts` now calls it on `updated` / `already_up_to_date` / `cache_refreshed`, so the modal won't re-fire on page reload after a successful update.

**P2 — hook self-correction.** `use-update-notification.ts` now resets `availableUpdate` to `null` when a later poll returns `updateAvailable: false`, instead of holding stale state.

Added `web-ui/src/hooks/use-update-notification.test.tsx` (5 cases) covering the happy path, up-to-date, the new self-correction, `dismiss()`, and rejected queries. Local checks: `tsc --noEmit` clean across root + web-ui, biome clean, all 498 web-ui + 108 relevant root tests passing.

robinnewhouse:
@greptileai

robinnewhouse:
Both points addressed in `b7d286d7`:

**P1 — wrong install command for `pnpm dlx` / `yarn dlx` / `bunx`.** `buildUserFacingInstallCommand` now takes `updateTiming` and branches: shutdown-timing PNPM/YARN/BUN return `pnpm dlx kanban` / `yarn dlx kanban` / `bunx kanban` (re-run the launcher) instead of `pnpm add -g …` etc. (which would change the user's workflow into a global install).

**P2 — misleading restart message on `already_up_to_date`.** The dialog success state now only shows "Updated to Kanban X. Restart Kanban …" when `status === "updated"`. For `already_up_to_date` and `cache_refreshed` it falls back to `result.message` (e.g. "Kanban is already up to date (1.0.0).").

Added 3 tests in `test/runtime/update/auto-update.test.ts` covering pnpm-dlx, yarn-dlx, and bunx shutdown-timing paths. 31 update tests + 111 trpc/runtime tests + 498 web-ui tests all pass; tsc + biome clean.

robinnewhouse:
@greptileai

web-ui/src/App.tsx
```diff
@@ -79,6 +81,7 @@ import { useTerminalThemeColors } from "@/terminal/theme-colors";
 import type { BoardData } from "@/types";
 
 export default function App(): ReactElement {
+	const { availableUpdate, dismiss: dismissUpdateNotification } = useUpdateNotification();
```
BarreiroT: Generally, I think we should start to debloat the `App.tsx` component. We could extract all of these to a Provider or something, that takes care of fetching the hook and conditionally rendering the `UpdateAvailableDialog`.

WDYT?

robinnewhouse: Always in favor of de-bloating. Let me see what I can do.

src/server/runtime-server.ts
```diff
@@ -66,6 +71,8 @@ export interface CreateRuntimeServerDependencies {
 	) => DisposeTrackedWorkspaceResult;
 	collectProjectWorktreeTaskIdsForRemoval: (board: RuntimeWorkspaceStateResponse["board"]) => Set<string>;
 	pickDirectoryPathFromSystemDialog: () => string | null;
+	getUpdateStatus?: () => RuntimeUpdateStatusResponse;
+	runUpdateNow?: () => Promise<RuntimeRunUpdateResponse>;
```
BarreiroT: Why can these be undefined?

robinnewhouse:
@greptileai

src/server/runtime-server.ts
robinnewhouse: Thanks for catching. fixed.

src/update/update.ts
```diff
@@ -72,6 +72,51 @@ interface PendingShutdownAutoUpdate {
...
+		case UpdatePackageManager.YARN:
+			return updateTiming === "shutdown" ? `yarn dlx ${packageName}` : `yarn global add ${packageSpec}`;
+		case UpdatePackageManager.BUN:
+			return updateTiming === "shutdown" ? `bunx ${packageName}` : `bun add -g ${packageSpec}`;
+		case UpdatePackageManager.NPX:
+			return `npx ${packageName}`;
+		case UpdatePackageManager.NPM:
+		case UpdatePackageManager.LOCAL:
+		case UpdatePackageManager.UNKNOWN:
+			return `npm install -g ${packageSpec}`;
```
BarreiroT, line 103: Hmm, I would prefer not to autoupdate on `unknown`.  WDTY?

src/update/update.ts
```diff
@@ -72,6 +72,51 @@ interface PendingShutdownAutoUpdate {
...
+	 * "shutdown": a transient (npx/pnpm dlx/yarn dlx/bunx) cache refresh is scheduled for shutdown.
+	 *   The next launch will use the new version.
+	 */
+	updateTiming: "startup" | "shutdown";
+	/**
+	 * User-facing command the user can copy-paste to apply the update. For global installs this is
+	 * the install command (e.g. `npm install -g kanban@latest`). For transient runs (npx, pnpm dlx,
+	 * yarn dlx, bunx) it's just the rerun command (e.g. `npx kanban`) since the auto-update has
+	 * already cleared the relevant cache on shutdown.
+	 */
```
BarreiroT: Do we need this comment?

src/update/update.ts
```diff
@@ -72,6 +72,51 @@ interface PendingShutdownAutoUpdate {
...
 
+export interface PendingUpdateNotification {
+	currentVersion: string;
+	latestVersion: string;
+	/**
+	 * "startup": a global-install update was spawned in the background. The running process still
+	 *   runs the old version, so the user needs to restart to pick up the new version.
+	 * "shutdown": a transient (npx/pnpm dlx/yarn dlx/bunx) cache refresh is scheduled for shutdown.
+	 *   The next launch will use the new version.
+	 */
```
BarreiroT: Hmm, generally speaking, I think comments for type definitions are smelly

web-ui/src/components/update-available-dialog.tsx
```diff
@@ -0,0 +1,146 @@
...
+
+	const handleCopy = async (): Promise<void> => {
+		try {
+			await navigator.clipboard.writeText(installCommand);
+			setCopied(true);
+			window.setTimeout(() => {
+				setCopied(false);
+			}, 1500);
+		} catch {
+			// Clipboard write can fail on some browsers/permissions; the user can still select the text.
```
BarreiroT: Do we wanna show a toast for them to copy it manually?

web-ui/src/components/update-available-dialog.tsx
```diff
@@ -0,0 +1,146 @@
...
+	};
+
+	const allowClose = phase.kind !== "running";
+
+	return (
+		<Dialog
+			open={open}
+			onOpenChange={(next) => {
+				if (!next && allowClose) {
+					onClose();
```
BarreiroT, line 78: Hmm, bit of a nit, it's weird to lock users here

web-ui/src/hooks/use-update-notification.test.tsx
```diff
@@ -0,0 +1,160 @@
...
+}
+
+async function importHookModule() {
+	const fetchRuntimeUpdateStatusMock = vi.fn<(workspaceId: string | null) => Promise<RuntimeUpdateStatusResponse>>();
+	vi.resetModules();
+	vi.doMock("@/runtime/runtime-config-query", () => ({
+		fetchRuntimeUpdateStatus: fetchRuntimeUpdateStatusMock,
+	}));
+	const module = await import("@/hooks/use-update-notification");
+	return { module, fetchRuntimeUpdateStatusMock };
```
BarreiroT: Why do we need this function? Mocks are hoisted, so they will always run before imports

web-ui/src/hooks/use-update-notification.ts
```diff
@@ -0,0 +1,83 @@
...
+				// Update status is a best-effort nudge; ignore failures.
+			}
+		}
+
+		void checkOnce();
+
+		// The server's pending-update state is populated asynchronously after
+		// startup, so give it a couple of retries shortly after mount before
+		// falling back to the slow poll interval.
+		const earlyRetry = window.setTimeout(() => {
```
BarreiroT: Why use `window.setTimeout`/`window.clearTimeout` in this PR? We usually use the globals directly.

BarreiroT (dismissed):
Hey! Mostly LGTM, thank you! Added a few minor comments. Approving regardless because none of them are blocking

BarreiroT (approved):
LGTM - thank you!

PR #409, opened 2026-04-25
fix: auto load latest model catalog on startup

abeatrix:
Configure Kanban-started Cline sessions to load the latest public model catalog on init, load private models after auth, and avoid failing startup when catalog loading errors occur.

Also updates the session runtime test to assert the local runtime catalog defaults are passed to the SDK host.


BarreiroT (approved):
LGTM - thank you!

PR #412, opened 2026-04-25
Support local overrides for editable Cline providers

arafatkatze:
## Summary

Fixes editing built-in OpenAI-compatible Cline providers such as LiteLLM from the provider edit modal.

The modal previously routed every edit through `updateCustomProvider`, but SDK `updateLocalProvider` only updates providers that already exist in the local `models.json` registry. Built-in providers like `litellm` exist in the SDK catalog, not in that local registry, so saving the edit surfaced `provider "litellm" does not exist`.

This PR makes first edit of an eligible built-in provider create a local override for the same provider ID before calling the SDK update path, so the SDK still owns model source fetching, settings persistence, and provider registration without depending on SDK error message text. It also makes the modal reload saved local override fields, including capabilities, instead of falling back to SDK catalog defaults.

## Changes

- Expose provider `client`, capabilities, local/custom status, `modelsSourceUrl`, timeout, and custom headers through the runtime provider catalog.
- Show the provider Edit button for non-OAuth OpenAI-compatible providers, including LiteLLM, while keeping managed OAuth providers like Cline out of this PR.
- Seed a local provider registry entry before SDK update when an eligible OpenAI-compatible built-in has no local entry yet.
- Roll back the seeded local override if the SDK update fails after seeding.
- Preserve SDK catalog `modelsSourceUrl` values when no local override exists, while allowing local overrides to clear or replace that field.
- Ensure the local models registry directory exists before writing `models.json`.
- Preload saved modal fields when reopening the edit modal so model source URL, timeout, custom headers, and capabilities round-trip correctly.
- Preserve explicit empty capability overrides instead of replacing them with add-provider defaults.
- Add focused UI tests for editable provider visibility and saved modal field reload behavior.

## Validation

- `npm --prefix web-ui test -- cline-add-provider-dialog.test.tsx cline-setup-section.test.tsx`
- `npm --prefix web-ui run typecheck`
- `npm --prefix web-ui run build`
- `git diff --check`

Manual UI test:

- Ran patched full stack app with isolated `CLINE_DATA_DIR`.
- Opened Settings → Cline provider settings.
- Selected LiteLLM and confirmed the Edit button appears.
- Opened the edit modal, changed provider fields, clicked Update provider.
- Confirmed the modal closed without `provider "litellm" does not exist` or `ENOENT` errors.
- Reopened Edit and confirmed saved timeout/custom header values reloaded into the modal.
- Toggled `vision`, clicked Update provider, reopened Edit, and confirmed `vision` stayed selected.
- Toggled `vision` back off, clicked Update provider, reopened Edit, and confirmed all capability toggles stayed off.

Note: root `npm run typecheck` is currently blocked in this local dependency setup by existing `@clinebot/core` export mismatches unrelated to this PR.

### DIFF
--- a/src/cline-sdk/cline-provider-service.ts
+++ b/src/cline-sdk/cline-provider-service.ts
@@ -757,6 +757,8 @@ export function createClineProviderService() {
 							env: provider.env,
 							capabilities: provider.capabilities,
 							modelsSourceUrl: provider.modelsSourceUrl?.trim() || null,
+							headers: provider.headers,
+							timeoutMs: provider.timeoutMs,
 						}))
 						.sort((left, right) => {
 							if (left.id === "cline") {
--- a/src/cline-sdk/cline-provider-service.ts
+++ b/src/cline-sdk/cline-provider-service.ts
@@ -747,12 +747,16 @@ export function createClineProviderService() {
 							id: provider.id,
 							name: provider.name,
 							oauthSupported: (provider.capabilities ?? []).includes("oauth"),
+							custom: provider.custom,
+							client: provider.client,
 							enabled:
 								selectedProviderId.length > 0 ? selectedProviderId === provider.id : provider.id === "cline",
 							defaultModelId: provider.defaultModelId ?? null,
 							baseUrl: provider.baseUrl?.trim() || null,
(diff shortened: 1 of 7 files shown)


arafatkatze:
@greptileai review the pr again

PR #428, opened 2026-04-28
Optimize the workspace metadata monitor

BarreiroT:
This PR uses the data we obtain from the probe to prevent getting the diff totals. This reduces the function execution time from `11.534ms` to `0.001ms`, which is mostly the case when polling with no changes yet.

### DIFF
--- a/src/server/workspace-metadata-monitor.ts
+++ b/src/server/workspace-metadata-monitor.ts
@@ -7,7 +7,12 @@ import type {
 import { getGitSyncSummary, probeGitWorkspaceState } from "../workspace/git-sync";
 import { getTaskWorkspacePathInfo } from "../workspace/task-worktree";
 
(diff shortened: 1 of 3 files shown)


BarreiroT:
@greptileai

BarreiroT:
@greptileai

saoudrizwan:
One freshness concern with the new polling behavior: when a task finishes, users may be looking at the kanban board and see the card move from in progress to review, but the file/line change counts on the card can still reflect the previous metadata snapshot until the 10s poll catches up. That could be confusing because the task appears ready for review before the visible git status has caught up.

Could we trigger an immediate workspace metadata refresh when a task session transitions out of running, or specifically when it enters review? That should keep the idle optimization while preserving immediate card status freshness at the point users are most likely to look at it. It should also bump taskWorkspaceStateVersion, which helps the open detail diff refetch without waiting for the slower poll.

PR #440, opened 2026-05-01
feat(desktop): auto-update the Kanban runtime in the background

### LINKED ISSUE #438: feat(desktop): keep desktop shell on latest kanban runtime (npm-fetch) + electron-updater for shell
## Background

#397 has landed the Electron desktop shell with the Kanban runtime **bundled at dmg build time** via `packages/desktop/scripts/stage-cli.mjs`. That gives us a reliable offline baseline, but it also means:

- The runtime served to desktop users is frozen at whichever root `kanban` version was staged when the dmg was built.
- The existing in-app update modal is CLI/npm-oriented (`kanban update`, `npm install -g`, `npx`, etc.) and does not correctly update a packaged `.app`.
- `packages/desktop/electron-builder.yml` has no `publish:` block, so the Electron shell itself cannot auto-update either.

For desktop to feel Cursor/VS Code-like (“new update available”, “restart to apply”) **and** to “serve the latest Kanban” quickly after npm publishes, we need two independent update channels:

1. **Runtime update channel** — fetch/install the latest `kanban` runtime from npm into user data, then use it on next app/runtime restart.
2. **Shell update channel** — update the Electron app/DMG itself for native deps, Electron, packaging, and `main.ts`/shell code.

> Review-boundary note: this is intentionally post-#397. #397 established the Electron shell + bundled-runtime fallback. Runtime self-update changes the security, integrity, rollback, dependency-resolution, and UX model enough to deserve a focused follow-up PR (or small PR series).

## Current packaged-runtime baseline

#397’s packaged runtime execution primitive is now in good shape for the bundled fallback:

- `packages/desktop/build/bin/kanban` prefers the bundled Electron binary with `ELECTRON_RUN_AS_NODE=1`, so a GUI-launched app does **not** require system `node` on PATH.
- It falls back to system `node` only for non-packaged/developer layouts.
- The bundled runtime is staged under `app.asar.unpacked/cli/cli.js` and launched by `RuntimeChildManager` as a subprocess.

The runtime updater should reuse this “known packaged runner” model. It should not depend on a user-installed Node/npm being available.

## Channel 1 — runtime npm fetch/store (fast iteration)

**Goal:** Kanban runtime/web-ui updates within minutes of `npm publish`, without requiring a dmg rebuild.

### Desired behavior

- App startup should start quickly from the current verified runtime, falling back to the bundled runtime when needed.
- In the background, desktop checks the selected npm dist-tag/channel (initially `latest`) for a newer `kanban` package.
- If a newer version is available, desktop downloads/installs it into a versioned runtime store under `${app.getPath("userData")}/kanban-runtime/`.
- The newly installed runtime is verified before any pointer swap.
- The update is applied on next app launch or explicit “Restart runtime to update”; do **not** hot-swap active task sessions.

### Proposed modules / responsibilities

Add a focused desktop runtime-update layer, likely split into small files rather than one god module:

- `packages/desktop/src/runtime-store.ts`
  - paths under `${userData}/kanban-runtime/`
  - `versions/<version>/`
  - `versions/<version>.partial/`
  - `current.json` pointer (prefer JSON over symlink for Windows friendliness)
  - bad-version markers
  - stale `.partial` cleanup and retention cap
- `packages/desktop/src/runtime-update-checker.ts`
  - npm registry metadata fetch
  - selected dist-tag/channel resolution (`latest` initially, future `stable`/`beta` possible)
  - semver comparison
  - bounded timeout (~5s)
- `packages/desktop/src/runtime-installer.ts`
  - download tarball
  - verify package identity/version/integrity using registry metadata (`dist.integrity`/shasum)
  - unpack to `.partial`
  - finalize atomically to `versions/<version>`
  - verify runnable entry and required assets before pointer update
- `packages/desktop/src/runtime-resolver.ts`
  - choose runtime in order: verified `current` → previous known-good → bundled fallback
  - expose runner/entry details to `RuntimeChildManager`

Names can change, but keep responsibilities separated enough to test pure store/check/install logic without booting Electron.

### Important constraints

- **Do not rely on system `node` or `npm`.** GUI app PATH is unreliable, and vanilla users may not have Node installed. Use the packaged Electron-as-Node runner (or another bundled runner) for downloaded runtimes too.
- **Do not shell out to `npm install` as the primary mechanism.** Prefer registry metadata + tarball download/extract so installs are deterministic and do not require npm on PATH.
- **Account for native/runtime dependencies.** The root npm package ships `dist/**`; `dist/cli.js` bundles most deps but currently externalizes `node-pty`. A downloaded runtime must have a deliberate strategy for `node-pty` resolution (for example: reuse the desktop-bundled compatible `node-pty`, stage/copy/link it into the runtime store, or make the root runtime bundle self-contained except for a clearly provided native layer). Do not accidentally rely on Node module resolution finding packaged `app.asar` internals from `${userData}`.
- **Use atomic installs.** Install into `versions/<version>.partial`, verify, then finalize and update `current.json`. Never mutate the active runtime in place.
- **Do not kill active task sessions automatically.** Download in background; apply on next app restart or explicit runtime restart.
- **Timeouts and fallback:** ~5s registry check, ~60s install/download budget initially. Startup must continue with current/bundled runtime on timeout/failure.
- **Rollback:** if a downloaded runtime fails startup or health checks, mark it bad and automatically fall back to previous known-good or bundled runtime.
- **Security/integrity:** verify package name/version and registry integrity before pointer swap. Keep the selected channel explicit.
- **Disk hygiene:** remove stale `.partial` installs and cap retained versions.

### Startup flow sketch

1. Resolve runtime for this launch:
   - read `current.json`
   - ignore missing/corrupt/bad versions
   - prefer verified current user runtime
   - otherwise use previous known-good
   - otherwise use bundled fallback
2. Start the runtime immediately.
3. In background, check npm for a newer version.
4. If newer:
   - download + verify + install into `versions/<version>`
   - update `current.json` only after verification
   - notify UI: “New Kanban runtime downloaded. Restart runtime to apply.”
5. On next launch/restart, use the new `current` runtime.
6. If startup health fails for a user runtime, mark bad and retry with previous/bundled fallback.

### UX

- Runtime update UI should be distinct from shell update UI:
  - Runtime update: “Restart runtime to apply” (lighter-weight; may affect active task sessions, so make explicit).
  - Shell update: “Restart app to update” / “Install and relaunch”.
- First connected launch can optionally show a bounded “Downloading newer Kanban runtime…” splash, but the default should favor fast startup with bundled/current runtime and background update.
- The existing web update modal currently says “Run this npm command” and calls `runtime.runUpdateNow`; that should remain for CLI-launched Kanban or be gated by update kind. Do not show npm-global instructions inside packaged desktop.

### Tests

- Pure unit tests:
  - dist-tag/channel resolution
  - semver/version comparison
  - runtime-store path construction
  - pointer read/write and corrupt pointer handling
  - partial-install cleanup
  - atomic rollover behavior
  - bad-version marking
  - fallback selection
  - integrity/package identity verification
- Integration-ish tests:
  - stale bundled + fresh npm metadata installs downloaded runtime and selects it after restart
  - offline path boots current/bundled runtime without errors
  - failed/corrupt install leaves `current.json` unchanged
  - selected runtime startup failure rolls back automatically
  - registry/download timeouts do not block UI startup indefinitely
  - downloaded runtime can resolve required native deps (`node-pty`) under packaged runner assumptions

## Channel 2 — electron-updater for shell + native deps

**Goal:** installed desktop apps update when Electron shell code, native deps, packaging, notarization/signing, or `main.ts` behavior changes.

### Implementation

- Add `electron-updater` to `packages/desktop/package.json`.
- Add a small desktop updater module around `autoUpdater` rather than wiring everything inline in `main.ts`.
- Gate update checks by `app.isPackaged`.
- Use GitHub Releases as the provider:
  ```yaml
  publish:
    - provider: github
      owner: cline
      repo: kanban
  ```
- macOS builder targets should include `zip` as well as `dmg`; the zip target is needed for `latest-mac.yml` update metadata.
- CI release flow must upload the `.dmg`, `.zip`, and `latest-mac.yml` assets to the GitHub Release.
- Code signing/notarization must be present for production macOS updates.

### UX

- Use native Electron update notifications/dialogs first, or add a separate desktop update surface in the renderer.
- Branches to handle:
  - no update available
  - update available/downloading
  - update downloaded → prompt restart/install
  - update error → non-blocking warning/log

### Tests

- Mock updater feed/events.
- Verify packaged-only gating.
- Verify “update downloaded → restart prompt” and “error/no update” branches.
- Verify release workflow produces/upload expected update metadata.

## Why two channels?

- Runtime/web-ui npm publishes may happen frequently and should reach desktop quickly.
- DMG/app updates are heavier because signing/notarization/release artifacts are involved.
- Runtime channel keeps product behavior current with low friction.
- Shell channel covers changes that npm runtime fetch cannot safely update: Electron, native deps/ABI, process orchestration, packaging, permissions, updater code itself.

## Suggested implementation order

1. **Runtime store foundations:** paths, pointer file, partial cleanup, bad-version markers, fallback selection.
2. **Registry/check/install utilities:** npm metadata fetch, tarball download, integrity verification, atomic finalize.
3. **Runner/dependency strategy:** make downloaded runtimes executable with packaged Electron-as-Node and explicit native-dep resolution.
4. **Startup resolver wiring:** verified user runtime → previous known-good → bundled fallback.
5. **Background update + restart-to-apply UX.**
6. **Rollback on startup-health failure.**
7. **Channel 2 shell updater:** `electron-updater`, builder `publish`, mac zip metadata, release workflow.

Channel 2 can be a separate PR if Channel 1 gets large.

## Scope estimate

Likely more than a tiny patch: runtime store/check/install/rollback + tests + CI updater work. Prefer a small PR series over one very large PR.

## Acceptance criteria

- [ ] On a connected machine, killing and restarting the desktop app within minutes of an `npm publish` of the selected `kanban` update channel causes the new version to be served from `${userData}/kanban-runtime/current.json`.
- [ ] On an offline machine, the desktop app starts and serves bundled or last-known-good runtime without errors.
- [ ] A failed/corrupt/timed-out runtime install never mutates the current pointer.
- [ ] A downloaded runtime that fails startup health checks is marked bad and rolls back automatically.
- [ ] Runtime update activation never kills active task sessions without explicit restart/app restart.
- [ ] Packaged desktop never requires system `node` or `npm` for either bundled or downloaded runtimes.
- [ ] Downloaded runtimes have an explicit, tested native dependency strategy for `node-pty`.
- [ ] When a new signed desktop release is published with updater metadata, an installed desktop app prompts the user to install the shell update.
- [ ] Unit + integration tests cover both update paths and failure modes.

cc @johnwschoi

johnwschoi:
Closes #438.

## Requirement

> Installed desktop shell should fetch / use the latest Kanban runtime **without requiring a shell reinstall**.

The shell currently freezes its bundled `cli.js` at desktop-build time. An installed shell can't pick up a newer runtime — users have to reinstall the desktop app to get runtime fixes. This PR closes that gap.

## What this PR does

| # | Requirement | Implementation |
|---|---|---|
| 1 | Installed shell checks npm for latest kanban runtime | `createRuntimeAutoUpdate` (in `runtime-auto-update.ts`) calls `pacote.manifest("kanban@latest")` 30s after first boot and every 30 min thereafter. Single-flight gated. |
| 2 | Downloads / stages the runtime safely | `checkAndStageLatestRuntime` (in `runtime-update.ts`) uses `pacote.extract` into `${userData}/runtime-store/versions/<v>.partial/`, copies bundled `node-pty` from `app.asar.unpacked/node_modules/`, then atomic-renames to `<v>/`. Crash-resilient — partial dirs are swept before each extract. |
| 3 | Uses it on next app/runtime restart | Pointer file at `${userData}/runtime-store/current.json` is read on every spawn by `loadOverride`; the orchestrator spawns the pointed-to `cli.js` instead of the bundled one. Renderer gets `runtime:update-staged` IPC for a "Restart to apply" banner. |
| 4 | Keeps bundled runtime as fallback | Two-arm resolver: **pointer → bundled**, with the effective launch version always `max(pointer, bundled)` so a stale older pointer can never override a newer bundled runtime after a shell upgrade. If the pointed-to version fails its startup probe, `onCliEntryOverrideFailed` marks the version bad (`bad-versions.json`), clears the pointer, removes the version dir, and the orchestrator immediately retries the same launch with the bundled cli. |
| 5 | No electron-updater / shell update pipeline | No DMG/ZIP/EXE/AppImage release matrix, no `latest-mac.yml` / `latest.yml` hosting, no signed installer pipeline. The shell binary on disk is never modified — only the runtime in `userData/`. |
| 6 | Uses pacote / npm-maintained tooling | `pacote.manifest()` for registry checks, `pacote.extract()` for tarball staging. No custom HTTP / tarball / integrity code. |

## Code changes

The whole feature is **3 new modules + small wiring in `main.ts`**. Architecture is intentionally functional — no lifecycle classes, no EventEmitters, no DI containers.

### New modules (`packages/desktop/src/`)

- **`runtime-store.ts`** (170 LOC) — on-disk layout under `${userData}/runtime-store/`:
  - Atomic pointer (`current.json` via `*.tmp`+rename) with strict validation: `semver.valid(version)` required, `cliEntry` must equal the canonical `versions/<v>/dist/cli.js` path after `path.resolve()` (any drift → pointer rejected at `readPointer`, and `writePointer` throws). This is what stops a tampered `current.json` from forwarding an arbitrary on-disk path to the shim as `KANBAN_CLI_OVERRIDE`.
  - `resolvePointerCliEntry` — returns the canonical cliEntry **only if the file exists on disk** (`statSync().isFile()`), so a wiped version dir can't permanently freeze the version gate.
  - `bad-versions.json` — version-blocklist that prevents re-staging a version that already failed startup. Entries are not pruned; the registry only publishes monotonically increasing versions and `isBadVersion(latest)` is the only call site, so old entries are dead weight (a few bytes) but never re-examined.
  - `cleanupPartials`, `removeVersionDir` — disk hygiene called on boot and after rollback.

- **`runtime-update.ts`** (99 LOC) — pure async function `checkAndStageLatestRuntime(opts) → StageOutcome`:
  - Discriminated outcome: `{ kind: "staged" | "up-to-date" | "already-staged" | "bad-version" }`.
  - Pacote `manifest` → version gate (against `currentVersion = max(pointer, bundled)`) → bad-version skip → already-staged skip → `pacote.extract` → stage `node-pty` from bundled → verify `dist/cli.js` exists → atomic rename partial → version dir → atomic pointer write. Any failure before the pointer write leaves the existing pointer untouched.

- **`runtime-auto-update.ts`** (198 LOC) — wires the two modules above into the orchestrator's `cliEntryOverride` callbacks and the 30s/30min check schedule:
  - `loadOverride()` enforces `max(pointer, bundled)`: a pointer at-or-below bundled (e.g. user upgraded the shell while userData still pointed at an older staged runtime) is treated as stale and cleared. Also clears pointers whose `cliEntry` no longer exists.
  - `onCliEntryOverrideFailed(reason)` is the rollback path: mark bad → remove version dir → clear pointer → broadcast `runtime:rolled-back` to the renderer.
  - `runCheck()` is single-flight gated, swallows network errors so the timer keeps firing, and broadcasts `runtime:update-staged` only on a successful new staging.
  - Returns `null` when `!isPackaged` so dev runs the bundled cli with no auto-update path active.

### Wiring (`packages/desktop/src/main.ts`, `runtime-orchestrator.ts`)

- **`main.ts`** — calls `createRuntimeAutoUpdate(deps)` once, passes its `resolveCliEntryOverride` / `onCliEntryOverrideFailed` callbacks into the orchestrator, and calls `scheduleChecks()` after the window is ready. Total wiring: +39 LOC.
- **`runtime-orchestrator.ts`** — added two optional callbacks: `resolveCliEntryOverride` (called on every spawn) and `onCliEntryOverrideFailed` (called when the staged cli fails its readiness probe; orchestrator then retries this same launch with the bundled cli, exactly once).

### File breakdown

| File | LOC | Role |
|---|---|---|
| `packages/desktop/src/runtime-store.ts` | 170 | **NEW** — disk layout, pointer, bad-versions list |
| `packages/desktop/src/runtime-update.ts` | 99 | **NEW** — `checkAndStageLatestRuntime` |
| `packages/desktop/src/runtime-auto-update.ts` | 198 | **NEW** — orchestrator wiring + check schedule |
| `packages/desktop/src/main.ts` | +39 | Calls `createRuntimeAutoUpdate(...)`, forwards callbacks |
| `packages/desktop/src/runtime-orchestrator.ts` | +64 | `resolveCliEntryOverride` / `onCliEntryOverrideFailed` hooks |
| `packages/desktop/src/preload.ts` | +33 | `runtime:update-staged` / `runtime:rolled-back` IPC bridge |
| `packages/desktop/test/runtime-store.test.ts` | 215 | **NEW** — pointer validation, canonical-path enforcement, bad-versions |
| `packages/desktop/test/runtime-update.test.ts` | 206 | **NEW** — pacote-mocked, all 4 outcomes + failure modes |
| `packages/desktop/test/runtime-auto-update.test.ts` | 290 | **NEW** — `loadOverride` (incl. stale-pointer + non-canonical), rollback, runCheck broadcast/error/stop |
| `packages/desktop/test/runtime-orchestrator.test.ts` | +265 | Override-callback wiring + fallback-on-probe-failure |
| `packages/desktop/test/runtime-child-manager.test.ts` | +44 | Cli-entry-override integration |

## Note on `node-pty` runtime closure

`node-pty/package.json` declares `node-addon-api` as a dependency, but it isn't required at runtime — node-addon-api is a header-only C++ helper consumed by `binding.gyp` during a from-source compile, and node-pty's prebuilt `.node` binaries (in `node-pty/prebuilds/`) are self-contained. Empirically verified: a `node_modules/` containing only `node-pty/` is sufficient for `require('node-pty').spawn(...)` to work. If a future kanban release grows a new external runtime dep beyond `node-pty`, `runtime-update` would still extract and rename it, then the new version would fail its startup probe, hit the rollback path, and end up on the bad-versions list — the user lands back on the bundled runtime automatically.

## Bad-version policy (user-visible behavior)

If `kanban@latest` fails startup on a user's shell once, this PR's bad-versions list will skip that exact version on every subsequent update tick. As soon as upstream publishes a newer version, the next check sees `latest > badVersion` and stages the new one normally (the old bad-versions entry is harmless dead data, never re-examined). Net effect from the user's perspective: one prompt to restart, then it falls back to bundled and stops nagging.

## Why not the "just stage at build time" alternative

Reviewer briefly suggested this. It satisfies the install-time slice but freezes the runtime until the next desktop reinstall — exactly what the requirement ("*without* shell reinstall") rules out. Build-time staging is necessary (and already present on `main` via `scripts/stage-cli.mjs`) but not sufficient.

## Out of scope

- Self-updating the **shell binary** (electron-updater + signed installer matrix). Requirement #5 explicitly excludes it.
- Channels / pre-release tracks. Single `@latest` only.
- Engines-compatibility and dependency-closure pre-extract gates. The rollback path covers both: an incompatible runtime fails its startup probe, gets marked bad, and the user lands on bundled.

## Tests + checks

- 291/291 desktop suite ✅ (33 new `runtime-store` / `runtime-auto-update` tests, 9 `runtime-update` tests; rest are orchestrator/child-manager wiring)
- `tsc --noEmit` clean
- Repo-wide pre-commit: 541/541 tests passing

Module-level coverage:

- `runtime-store` — pointer round-trip + atomic-write, strict semver + canonical-path validation on both `readPointer` (rejects non-canonical) and `writePointer` (throws on non-canonical), `resolvePointerCliEntry` self-repair contract, bad-versions idempotency / semver-sort / file-corruption defense, `cleanupPartials`.
- `runtime-update` — pacote mocked at module boundary; all 4 outcomes (`staged` / `up-to-date` / `already-staged` / `bad-version`) + non-semver `currentVersion` defense, missing `node-pty`, corrupt tarball (no `dist/cli.js`), stale-partial recovery.
- `runtime-auto-update` — `loadOverride` (stale-pointer-clears, file-missing self-repair, non-canonical pointer, returns canonical path on happy path), `onCliEntryOverrideFailed` rollback sequence, `runCheck` broadcast on staged + uses `max(pointer, bundled)` as `currentVersion` + swallows pacote/network errors + `stop()` cancels timers.
- `runtime-orchestrator` — `resolveCliEntryOverride` wiring, `onCliEntryOverrideFailed` invocation with bundled-cli retry, no-double-failure when bundled-arm itself fails, `resolveCliEntryOverride` throwing doesn't brick the spawn.


johnwschoi:
Closing on reviewer feedback: the stated requirement ("the desktop shell ships and runs the latest Kanban runtime") is already satisfied on `main` by `packages/desktop/scripts/stage-cli.mjs`, which copies the freshly-built root `dist/cli.js` into the desktop bundle at every build, and `main.ts`'s existing `resolveCliShimPath` plumbing that launches it.

This PR adds runtime OTA (background npm polling, user-data runtime store, pointer file, rollback subsystem, pacote, restart-to-apply IPC, native dep copying into userData) on top of that already-working baseline. None of that is requested. If we ever decide to support runtime OTA as an explicit feature, we can revisit; until then the packaging-time fix on main is the right shape.

Closes #438 will need to stay open or be reframed against the actual gap (if any) — given main already stages the latest runtime, #438 may already be implicitly resolved; flagging for triage.

johnwschoi:
Reopening: clarified requirement is **"installed desktop shell fetches/uses latest runtime without requiring shell reinstall."** That's runtime OTA, not packaging-time staging. The bundled-runtime-at-build-time path on `main` satisfies only the install-time slice; after that the runtime stays frozen until the user reinstalls the shell, which is exactly the gap this PR closes.

The earlier review feedback ("reduce to packaging-only") was based on a narrower reading of the requirement than the user actually has. The current branch is the trimmed version (275 tests, +3377 / -177 LOC, all small functional modules — no over-engineering). Sticking with this shape.

PR #446, opened 2026-05-05
chore: Cline SDK 0.0.38 

abeatrix:
Update kanban sdk usage to match the Cline SDK 0.0.38 interface. Removes llms, agents packages from dependencies as they are now exposed by the core package.
Update to use the new hook names.


abeatrix:
@greptileai

saoudrizwan:
Pushed 5441d1e8 to keep the chat panel on the previous message-event contract while preserving the SDK 0.0.38 adapter work.

The concern was that `latestHookActivity.finalMessage` is summary metadata. It is intentionally normalized for compact status/card display, so using it as temporary chat transcript text can flatten markdown/code formatting and can also fail prefix matching against real streamed assistant messages. The safer shape is the old boundary: the SDK adapter translates all new assistant event forms into raw chat messages, and the web chat panel renders only those messages.

I removed the summary-derived assistant preview from `useClineChatPanelController` and the tests that asserted it. The new SDK paths still emit real assistant messages through the runtime adapter, including `assistant-text-delta`, `assistant-message`, `content_start`, `content_end`, and final run output cases.

saoudrizwan (approved):

BarreiroT (approved):
LGTM - Thank you!

PR #449, opened 2026-05-06
Fix template injection

BarreiroT:
Env variables injected into the `run` can lead to a template injection. This PR prevents it.


TheRealSpencer (approved):
Nice and simple!

PR #479, opened 2026-05-11
Fix Windows npm shim PTY launches

anst-dev:


anst-dev:
Addressed both review comments:
- Made `%dp0%` path replacement case-insensitive to match the shim parsing regex.
- Added `canAccessPath` guard before directly launching resolved `.exe/.com` targets.

Added regression tests for mixed-case `%DP0%` and missing executable fallback.

Validation:
- `npm run test -- test/runtime/core/windows-cmd-launch.test.ts test/runtime/terminal/pty-session.test.ts`
- `npm run typecheck`
- `npx @biomejs/biome check src/core/windows-cmd-launch.ts src/terminal/pty-session.ts test/runtime/core/windows-cmd-launch.test.ts test/runtime/terminal/pty-session.test.ts`

PR #497, opened 2026-05-18
fix(cli): --no-passcode flag properly disables passcode authentication

RytisLT:
## Problem

The `--no-passcode` CLI flag was not working - the server still required a passcode even when the flag was passed.

## Root Cause

Commander.js `--no-*` flags set the base property to `false` (not `true`). The code was checking `options.noPasscode === true`, which never matched.

## Fix

Align `--no-passcode` handling with the existing `--no-open` pattern:
- Changed `RootCommandOptions` to use `passcode?: boolean` (matching how Commander stores `--no-*` flags).
- Updated the CLI option mapping to read `options.passcode === false` instead of `options.noPasscode === true`.


simonsystem:
@RytisLT: Thank for providing this MR. I would choose this one for merge, because its provided with some tests. **Maintainer**: Is there something we can do, to speedup merging?

PR #505, opened 2026-05-20
llama.cpp context overflow response fix

### LINKED ISSUE #504: Context Overflow Compaction Fails with llama.cpp
### What happened?

Kanban Agent and Individual tasks failed when hitting the context limit of my local llama.cpp server:
request (60179 tokens) exceeds the available context size (60160 tokens), try increasing it
^^ This response comes from llama.cpp

### Steps to reproduce

1. Launch llama.ccp "llama-server (your-model) -c 60000
2. Launch Kanban and configure provider to point there (I've done this with both the OpenAI Compatible/Native provider and the Ollama provider)
3. Run the agent ragged until you hit your context limit :)

### Expected behavior

Context compaction per the cline-context-overflow-compaction.ts in cline-sdk

### Kanban version

0.1.68

### Environment

WSL2 Ubuntu 24.04 LTS

John-Ridgway:
fix(cline-sdk): added 2 additional entries to CONTEXT_OVERFLOW_ERROR_PATTERNS to handle llama.cpp context overflow response patterns

fixes #504

### DIFF
--- a/src/cline-sdk/cline-context-overflow-compaction.ts
+++ b/src/cline-sdk/cline-context-overflow-compaction.ts
@@ -30,6 +30,8 @@ const CONTEXT_OVERFLOW_ERROR_PATTERNS = [
 	/reduce.*length.*messages.*completion/i,
 	/tokens?\s*>\s*[\d,]+\s*(maximum|limit)/i,
 	/input tokens?.*(exceed|exceeds).*(limit|maximum|context)/i,
+	/available context size/i,
+	/exceeds?\s*the available context size/i,
 ];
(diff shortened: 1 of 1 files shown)


John-Ridgway:
Discovered that it's deeper than just the context error array. llama.cpp sends the response like a regular message with finishReason "error" and text "request (60179 tokens) exceeds the available context size (60160 tokens), try increasing it"

I'm working on manually catching these "errors" and will re-submit the PR

John-Ridgway:
This will have to be handled at a much lower level since the "truncated" flag accessible in the raw llama cpp output is what needs targeting to accommodate all edge cases. Cancelling PR since it's likely I need to target a dependency's repo.

PR #516, opened 2026-05-25
fix: virtualize cline chat message list with react-virtuoso

### LINKED ISSUE #515: `cline-agent-chat-panel`: UI becomes unusable in long sessions due to unvirtualized message list
### What happened?

In sessions with many accumulated chat messages, the browser UI slows to a crawl and eventually becomes unresponsive. `cline-agent-chat-panel.tsx` renders every message as a live DOM node with no virtualization. Layout and paint cost scales linearly with total message history rather than the visible viewport, so long-running agent sessions progressively degrade until the app is unusable.

### Steps to reproduce

1. Run one or more agent tasks through a long session (hundreds of messages accumulated)
2. Switch between task cards and observe chat panel lag
3. Scroll the message list while streaming is active

Scrolling, switching tasks, and general interaction become progressively sluggish. With enough history the app becomes effectively unusable.

### Expected behavior

The UI remains responsive regardless of message history length.

### Kanban version

0.1.68

### Environment

All browsers and all OS

nzneit:
Replaces the full-DOM message list in cline-agent-chat-panel with react-virtuoso so long sessions no longer render every message as a live DOM node. Also wraps ClineChatMessageItem in React.memo so unchanged messages skip re-renders during streaming.

Resolves: https://github.com/cline/kanban/issues/515

### DIFF
--- a/web-ui/src/components/detail-panels/cline-agent-chat-panel.test.tsx
+++ b/web-ui/src/components/detail-panels/cline-agent-chat-panel.test.tsx
@@ -8,6 +8,31 @@ import type { ClineChatMessage } from "@/hooks/use-cline-chat-session";
 import type { RuntimeTaskHookActivity, RuntimeTaskSessionSummary } from "@/runtime/types";
 import { resetWorkspaceMetadataStore, setTaskWorkspaceSnapshot } from "@/stores/workspace-metadata-store";
 
+vi.mock("react-virtuoso", () => ({
+	Virtuoso: ({
+		data,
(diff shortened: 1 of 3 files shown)


DevinJamesK:
Awesome update, I will test this ASAP.

TyroneNel:
Re-implemented the Virtuoso chat list on current main as https://github.com/cline/kanban/pull/599. The May draft no longer matches the panel (controller hook, model picker, images).

PR #557, opened 2026-07-05
fix: --no-passcode commander option not working

Sureiya:
The --no-passcode flag was ignored because Commander creates options.passcode (negated from --no-passcode) but the code checked options.noPasscode which is never set.


RytisLT:
see https://github.com/cline/kanban/pull/497

simonsystem:
@Sureiya: RytisLT ist right, there is already a PR for it. Please close this one, to keep the kanban project clean of PRs.

PR #580, opened 2026-08-05
fix(cline-sdk): unblock ClinePass cards and model selection

### LINKED ISSUE #552: Broken ClinePass provider
### What happened?

Unable to set/use cline-pass provider in kanban settings.
Able to select the cline-pass provider, but weirdly shows only one model at a time if at all.
Shows this error when prompting a model set with cline-pass: Unknown or disabled provider "cline-pass". 

Image

Image

### Steps to reproduce

1. configure cline cli with clinepass
2. open cline kanban
3. Go to settings -> Cline -> Api provider
4. Try to set it to cline-pass and try to look for all models clinepass gives

### Expected behavior

Should show and be able to select all models clinepass gives. 

### Kanban version

kanban@0.1.68

### Environment

MacOS Version 26.5.1 (25F80)

DanBeckDev:
fixes #552

This is a fix for the `bug`-labeled report in #552 — ClinePass cards fail to start, and
the model picker shows a single model. Feature-request discussion #547 has the background
and the original root-cause analysis.

## The bug

Starting a card with the `cline-pass` provider fails immediately:

```
Error: Unknown or disabled provider "cline-pass".
```

Settings also lets you *select* ClinePass but shows only a single model — both symptoms in #552.

## Root cause

The bundled `@clinebot/core` provider registry has no `cline-pass` entry, so `GatewayRegistry.resolveModel` throws before any request is made. Reproducible against the pinned SDK in three lines:

```js
import { Llms } from "@clinebot/core"; // 0.0.38, what kanban@0.1.70 bundles
const gw = Llms.createGateway({ providerConfigs: [{ providerId: "cline-pass", apiKey: "k" }] });
await gw.stream({ providerId: "cline-pass", modelId: "cline-pass/glm-5.2", messages: [{ role: "user", content: "hi" }] });
// → Unknown or disabled provider "cline-pass".
```

The two UI symptoms are separate Kanban-side effects of the same gap:

- `getProviderCatalog()` synthesizes an entry for whatever provider the shared `~/.cline` config has selected, so `cline-pass` looks supported even though the SDK has never heard of it.
- `getProviderModels()` falls back to the single saved model when the SDK returns none — hence "only one model at a time".

Bumping the dependency does not help: `@clinebot/core@latest` on npm is still `0.0.38`, and no version in that lineage defines `cline-pass`. The provider only exists in the newer `@cline/core` + `@cline/llms` lineage that the `cline` CLI and the extension run on, and migrating Kanban across that gap (0.0.38 → 0.0.69, four packages) is a much larger change than this issue.

## The fix

ClinePass is not a separate API surface. Comparing the two provider definitions in `@cline/llms`, `cline-pass` and `cline` are identical apart from name, default model and model list:

| | `cline` | `cline-pass` |
|---|---|---|
| baseUrl | `https://api.cline.bot/api/v1` | `https://api.cline.bot/api/v1` |
| protocol / client | `openai-chat` / `openai-compatible` | `openai-chat` / `openai-compatible` |
| auth | Cline account OAuth (`workos:` token) | Cline account OAuth (`workos:` token) |
| env | `CLINE_API_KEY` | `CLINE_API_KEY` |

Only the model namespace differs — the Cline API bills a request against the subscription when the model id is `cline-pass/*`.

So Kanban keeps `cline-pass` as its own selectable provider (own catalog entry, own model list, own saved settings, which stay keyed by the id the CLI and extension write) and hands the bundled SDK the `cline` provider id when it starts a session, leaving the `cline-pass/*` model id untouched. Verified against the pinned SDK with an intercepting `fetch` — routing a `cline-pass/*` model through the registered `cline` provider produces exactly the request the newer SDK would send:

```
POST https://api.cline.bot/api/v1/chat/completions
Authorization: Bearer workos:<token>
{"model":"cline-pass/glm-5.2","stream":true,...}
```

Changes, all inside the `src/cline-sdk/` boundary plus two web-ui call sites:

- **`cline-pass-provider.ts`** (new) — provider ids, the `isClineAccountProviderId` predicate, and `resolveSdkRuntimeProviderId`, which maps `cline-pass` → `cline` for SDK calls only.
- **Session start and system prompt** — routed through the registered `cline` provider, so ClinePass gets the same workspace-metadata prompt the official CLI uses.
- **Catalog** — a ClinePass entry derived from the registered `cline` provider (no restated endpoint/capabilities), so it is offered even when it is not the current selection. It stands down automatically if the SDK ever registers `cline-pass` itself; there is a test for that.
- **Model list** — read from the public model catalog the SDK already uses for live model metadata (`DEFAULT_MODELS_CATALOG_URL`, which publishes the ClinePass models with names, capabilities and context windows), cached for 10 min in the service closure like the existing profile cache, and degrading to the saved model when offline. No hardcoded model table.
- **Cline account features** — OAuth login/refresh, `CLINE_API_KEY`, and profile / balance / organizations / account switching / Featurebase now accept both Cline-account providers instead of only `cline`. Previously, selecting ClinePass silently blanked all of them.

The one hardcoded value is `CLINE_PASS_DEFAULT_MODEL_ID = "cline-pass/glm-5.2"`, used as the catalog entry's default model; it mirrors the `defaultModelId` published for `cline-pass` in `@cline/llms`.

## Testing

- `test/runtime/cline-sdk/cline-pass-provider.test.ts` — provider-id predicates and SDK routing.
- `test/runtime/cline-sdk/cline-pass-provider-service.test.ts` — catalog contribution (including deferring to the SDK), model listing from the catalog plus the offline fallback, launch config (token prefixing, `CLINE_API_KEY`, ClinePass-named error, default model), and Cline account features with ClinePass selected.
- `test/runtime/cline-sdk/cline-session-runtime.test.ts` — regression test that a ClinePass session starts on `providerId: "cline"` while keeping `modelId: "cline-pass/glm-5.2"`.
- Existing `@clinebot/core` test mocks gained the `DEFAULT_MODELS_CATALOG_URL` export.

`npm run check` and `npm run build` both pass. `npm run test` matches the pre-existing failures on `main` (`git-history`, `runtime-state-stream.integration`, `task-command-exit.integration` — all load-sensitive and passing in isolation). `dist/cli.js` now contains the `cline-pass` handling that was previously absent.

## Notes for reviewers

- If a maintainer would rather source the model list from the authenticated `GET /api/v1/models` endpoint (the account's real entitlements) than from the public catalog, that swap is contained to one function — I used the public catalog because I could verify its shape without a ClinePass account.
- Credentials are read from the provider settings keyed by the selected provider id, with no fallback between `cline` and `cline-pass`. If a user's token happens to live only under `cline`, ClinePass shows as not signed in and one sign-in from Settings fixes it. Happy to add the fallback if you'd prefer it.
- I have a ClinePass-less environment, so the request shape is verified by interception against the pinned SDK rather than a live subscription call. A maintainer with a ClinePass account can confirm the round trip in one card.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/cline-sdk/cline-pass-provider.ts
+++ b/src/cline-sdk/cline-pass-provider.ts
@@ -1,0 +1,52 @@
+// ClinePass (`cline-pass`) identity helpers for the bundled Cline SDK.
+//
+// The bundled `@clinebot/core` provider registry has no `cline-pass` entry, so a
+// card that inherits `cline-pass` from the shared `~/.cline` config fails at
+// model resolution with `Unknown or disabled provider "cline-pass"` before any
+// request is made.
+//
+// ClinePass is not a separate API surface though: it is the same Cline endpoint,
+// the same Cline account OAuth token, and the same OpenAI-compatible protocol as
+// the usage-billing `cline` provider. Only the model namespace differs — the
+// Cline API bills a request against the subscription when the model id is
+// `cline-pass/*`.
+//
+// So Kanban keeps `cline-pass` as its own selectable provider (own catalog
+// entry, own model list, own saved settings) and hands the bundled SDK the
+// `cline` provider id when it starts a session, leaving the `cline-pass/*` model
+// id untouched. Once the bundled SDK registers `cline-pass` itself, its own
+// definition takes over and these helpers stop contributing anything.
+
+export const CLINE_PROVIDER_ID = "cline";
+export const CLINE_PASS_PROVIDER_ID = "cline-pass";
+export const CLINE_PASS_PROVIDER_NAME = "ClinePass";
+export const CLINE_PASS_BASE_URL = "https://api.cline.bot/api/v1";
+// Mirrors the `cline-pass` default model published by the newer `@cline/llms`
+// provider catalog that the Cline CLI and extension run on.
+export const CLINE_PASS_DEFAULT_MODEL_ID = "cline-pass/glm-5.2";
+
+export function isClinePassProviderId(providerId: string | null | undefined): boolean {
+	return providerId?.trim().toLowerCase() === CLINE_PASS_PROVIDER_ID;
+}
+
+/**
+ * True for every provider backed by a Cline account: usage-billing `cline` and
+ * subscription `cline-pass` share one OAuth token, one API base URL, and one set
+ * of account endpoints (profile, balance, organizations, Featurebase).
+ */
+export function isClineAccountProviderId(providerId: string | null | undefined): boolean {
+	const normalizedProviderId = providerId?.trim().toLowerCase();
+	return normalizedProviderId === CLINE_PROVIDER_ID || normalizedProviderId === CLINE_PASS_PROVIDER_ID;
+}
+
+/**
+ * Maps a Kanban provider id onto the provider id the bundled SDK knows.
+ * `cline-pass` requests run through the registered `cline` provider; the
+ * `cline-pass/*` model id is what routes them to the subscription.
+ *
+ * Saved provider settings stay keyed by the Kanban provider id, because that is
+ * the key the Cline CLI and extension write in the shared config.
+ */
+export function resolveSdkRuntimeProviderId(providerId: string): string {
+	return isClinePassProviderId(providerId) ? CLINE_PROVIDER_ID : providerId;
+}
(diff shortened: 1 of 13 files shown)


Risyandi:
I'm trying to mention @saoudrizwan for the review. Several users got the issue when using clinepass in kanban. thank you

aaronwng:
can we have this merged?

Issue #581, opened 2026-08-05
Web UI freezes ~60s (page unresponsive) when switching tasks during a long agent run (terminal snapshot + 10k scrollback on main thread)


bradrushworth:
# Web UI freezes ~60s ("page unresponsive") when switching tasks during a long agent run

**Package:** `kanban` (npm)
**Affected version:** `0.1.70` (latest release as of 2026-08-05; also the currently running instance)
**Component:** Web UI terminal (`xterm.js`) + server-side terminal mirror (`terminal-state-mirror.ts`) + `persistent-terminal-manager.ts`
**Severity:** High (makes the board unusable while agents run)
**Type:** Performance / main-thread block (browser + Node server)

---

## Summary

When an agent session is active and the user clicks between tasks in the browser, the
tab becomes unresponsive for **~60 seconds**, and during an active run the browser shows
"page is unresponsive" (main thread blocked). The freeze is dominated by the **terminal
restore/snapshot path**: on every task switch the server serializes a **full headless
xterm buffer (10,000 lines of scrollback)** and ships it over the WebSocket, and the
browser re-parses and re-renders the entire buffer in an xterm configured with the **same
10,000-line scrollback**. Long/complex runs grow this buffer (and the persisted transcript),
so cost scales with run length — which matches the user's report that "long/complex agent
runs filling context and producing huge logs" makes it worse.

This is **not** fixed by v0.1.70 (v0.1.70 only fixed the unrelated "clear completed tasks
on large boards" freeze, #564).

---

## Environment

- `kanban` `0.1.70` (confirmed via `kanban --version`, package `version`, and the live process)
- Running process: `node /usr/local/bin/kanban --host 127.0.0.1 --port 3484 --no-open` (PID 1)
  - RSS ≈ 590 MB, already elevated — consistent with the OOM trajectory in #273
  - Agent child processes (`.fork()` / `child_process` / `worker_threads`) are spawned **under this same process**, so CPU/IO contention is on the same host/event loop (shared-host contention).
- Client: Chromium/Chrome (any modern browser)

---

## Symptoms → root cause

| # | Symptom | Root cause (verified in source) |
|---|---------|---------------------------------|
| 1 | Clicking between tasks ≈ 60 s latency | Task switch opens a WebSocket to that task's terminal and the server sends a **full `restore` snapshot** of the headless terminal buffer; browser `applyRestore` does `terminal.reset()` + `terminal.write(snapshot)` of up to 10,000 lines, blocking the main thread. |
[...]

PR #582, opened 2026-08-06
fix(packaging): ship dist/entry.js hub-daemon spawn target (#495)

### LINKED ISSUE #495: `task start` hangs ~30s with `selectedAgentId=cline`: published tarball is missing `dist/entry.js`, hub daemon spawn fails
### What happened?

With `selectedAgentId=cline`, every attempt to start a task (CLI or web UI) hangs and the runtime eventually times out the underlying hub command after ~30 s. The `task list` command and the web UI itself are fine — only the agent‑launch step fails. After several attempts the runtime also runs out of memory and crashes (V8 OOM, 4 GB heap exhausted), so the failure isn't purely a hang — repeated retries leak something serious.

Symptom from the CLI (kanban@0.1.68, Node 22.22.2):

```
$ kanban task start --project-path /srv/minecraft --task-id 30fa2
Command failed: Command timed out after 30000ms
```

### Root cause

The `kanban` package’s `dist/cli.js` resolves the hub‑daemon entry as a **sibling of itself** in `dist/`:

```js
// dist/cli.js (paraphrased; deobfuscated from the bundle)
function OF() {
  let f = import.meta.url.endsWith(".ts") ? "ts" : "js";
  return fileURLToPath(new URL(`./entry.${f}`, import.meta.url));
}
function CF(cwd, opts) {
  let entry = OF();                       // → <pkg>/dist/entry.js
  let exec  = process.execPath;
  // …spawn `node <entry> --cwd <cwd> …`
}
```

But the **published npm tarball does not contain `dist/entry.js`**. Verified for every recent release from 0.1.50 through 0.1.68:

```
$ npm pack kanban@0.1.68 --dry-run
npm notice Tarball Contents
npm notice 16.4MB dist/cli.js
npm notice 28.6MB dist/cli.js.map
npm notice 566.3kB dist/index.js
npm notice 947.6kB dist/index.js.map
npm notice 2.2MB  dist/web-ui/assets/…
# …no dist/entry.js
```

So the daemon Node subprocess immediately crashes with `MODULE_NOT_FOUND`, e.g. from `~/.cline/data/logs/hub-daemon.log`:

```
node:internal/modules/cjs/loader:1386
  throw err;
  ^
Error: Cannot find module '/usr/.../node_modules/kanban/dist/entry.js'
    at Function._resolveFilename (node:internal/modules/cjs/loader:1383:15)
    …
  code: 'MODULE_NOT_FOUND'
```

The kanban runtime itself stays up and listening on `:3484` (so `task list` works), but it’s now stuck waiting on a hub it failed to spawn. `task start` ultimately bails with the built‑in hub‑command timeout:

> *“Hub command … timed out after 30000ms … Check hub‑daemon.log for matching command.start/command.slow entries, or run 'cline doctor fix' to restart the hub.”*

(error string is in the bundle at `LW = class extends Error { command; code; … }` / `Hub command ${name} timed out after ${ms}ms …`)

Naïvely copying `kanban/node_modules/@clinebot/core/dist/hub/daemon/entry.js` to `kanban/dist/entry.js` lets the require succeed but then crashes with `ReferenceError: AI is not defined` at `entry.js:182`. `AI` is a minified symbol that **is** defined in `dist/cli.js` (`function AI(…) {…}`), confirming that `dist/entry.js` was meant to be a kanban‑built bundle containing the same shared chunk — not a verbatim copy of the @clinebot/core file. The bundling step that produces it appears to be missing from the publish pipeline.

### Side effect: OOM after retries

When the broken state persists and the runtime is poked a few times (UI clicks, CLI retries), the kanban Node process runs out of memory:

```
<--- Last few GCs --->
[326674:0x4401d000]  1089040 ms: Mark-Compact 4092.3 (4122.5) → 4092.3 (4122.3) MB, …
FATAL ERROR: Reached heap limit Allocation failed - JavaScript heap out of memory
```

So the daemon‑spawn failure path likely keeps creating retry state / unflushed buffers somewhere.

### Steps to reproduce

1. Fresh global install of `kanban@0.1.68` and `cline@3.0.5`/`3.0.6` (both reproduce).
2. `~/.cline/kanban/config.json`:
   ```json
   { "selectedAgentId": "cline" }
   ```
3. Open the kanban web UI, create a task in backlog for any project.
4. Click **Start task** (or `kanban task start --project-path <repo> --task-id <id>`).
5. Observe: the start call hangs for ~30 s and errors with `Command timed out after 30000ms`. `~/.cline/data/logs/hub-daemon.log` shows the `MODULE_NOT_FOUND` above. The task’s worktree under `~/.cline/worktrees/<id>/...` *is* created successfully — failure is purely at the agent‑launch step.

### Workaround

Setting `selectedAgentId` to any other catalog entry that goes through `terminalManager.startTaskSession` instead of the cline hub path works fine (verified plumbing — `claude`, `codex`, `droid`, `kiro` all bypass `dist/entry.js`). The cline path is the only one that requires the missing file.

### Expected behavior

`kanban task start` with `selectedAgentId=cline` should launch the cline hub‑daemon successfully and start the task. Either:
- the publish pipeline should emit `dist/entry.js` (kanban‑bundled, self‑contained), or
- `OF()` should resolve the daemon entry from a path that actually exists in the published layout (e.g. `node_modules/@clinebot/core/dist/hub/daemon/entry.js`) and the entry there should be a runnable, self‑contained bundle.

### Kanban version

v0.1.68 (also tested back to 0.1.50 — `dist/entry.js` is missing from every published tarball I checked)

### Environment

- OS: Ubuntu 24.04.4 LTS, kernel 6.8.0
- Node: v22.22.2
- npm global prefix: `~/.npm-global`
- `cline` CLI: 3.0.5 → 3.0.6 (both reproduce)
- selected agent: `cline` (`~/.cline/kanban/config.json`)

bradrushworth:
## Summary

The `kanban` CLI spawns the hub daemon from `dist/entry.js`, but the build only emits `dist/cli.js` and `dist/index.js`. On a fresh install, every spawn attempt crashes with:

```
Error: Cannot find module /usr/local/lib/node_modules/kanban/dist/entry.js
```

This crash-loops ~75 times on startup (see `hub-daemon.log`), hanging the CLI and freezing the frontend. The real daemon implementation already ships inside the embedded `@clinebot/core` SDK, whose `exports` map exposes it as the public subpath `@clinebot/core/hub/daemon-entry` → `./dist/hub/daemon/entry.js`.

Fixes #495.

## Root cause

`scripts/build.mjs` bundles two entry points (`src/cli.ts` → `dist/cli.js`, `src/index.ts` → `dist/index.js`) but there is **no third entry point for the hub daemon spawn target**. The daemon spawn code (inlined from `@clinebot/core` into `dist/cli.js`) resolves its own location relative to the package root and expects `dist/entry.js` to exist — but nothing builds it, so it's absent from the published tarball.

## The fix

Add `src/entry.ts` as a thin re-export shim and a third `esbuild.build()` entry point in `scripts/build.mjs` so `dist/entry.js` is emitted and shipped in the tarball.

### Two non-obvious traps that the shim handles

1. **ESM constraint:** The `kanban` package declares `"type": "module"` (ESM). A CommonJS shim using `require()`/`module.exports` throws `ReferenceError: require is not defined in ES module scope` when Node loads it. The shim **must** use ESM `import`/`export` syntax.

2. **Exports-map gotcha:** The deep path `@clinebot/core/dist/hub/daemon/entry.js` is blocked by the `exports` field in `@clinebot/core/package.json`. The shim uses the **public exports-mapped subpath** `@clinebot/core/hub/daemon-entry` as the primary import, with a `import.meta.url`-relative fallback to the nested file for environments that don't honor the exports map.

## Changes

- **`src/entry.ts`** (new): ESM re-export shim. Uses top-level `await import("@clinebot/core/hub/daemon-entry")` with a fallback that resolves `node_modules/@clinebot/core/dist/hub/daemon/entry.js` relative to `import.meta.url` (since `__dirname` is not defined in ESM). Exports `default` and re-exports all named exports.
- **`scripts/build.mjs`**: Add a third `esbuild.build()` entry point (`src/entry.ts` → `dist/entry.js`) to the existing `Promise.all([...])`, and update the completion log message.

## Verification

This shim was validated against a live v0.1.70 install before porting to the build pipeline:
- `node dist/entry.js` loads the core daemon module with **no** `Cannot find module` error.
- `hub-daemon.log` goes to 0 bytes and stays clean — zero `Cannot find module .../kanban/dist/entry.js` entries.
- The CLI no longer hangs on startup and the frontend loads.

## Acceptance criteria

- [ ] After `npm run build`, `dist/entry.js` exists.
- [ ] `npm pack --dry-run` lists `dist/entry.js` in the tarball (`files` already includes `dist`).
- [ ] A fresh `npm install -g kanban` starts the hub daemon cleanly with no manual patching.
- [ ] `hub-daemon.log` contains zero `Cannot find module .../kanban/dist/entry.js` entries.

## Out of scope

- No changes to the minified `cli.js` spawn-target path (stays `dist/entry.js`).
- No changes to `@clinebot/core`'s `exports` map — `@clinebot/core/hub/daemon-entry` is the correct public entry.
- This is a `kanban` package packaging fix only; no runtime/workspace code changes beyond the new shim.

### DIFF
--- a/scripts/build.mjs
+++ b/scripts/build.mjs
@@ -28,7 +28,7 @@ const cjsShimBanner = [
 	"const require = __kanban_createRequire(import.meta.url);",
 ].join("\n");
 
-/** Shared esbuild options for both entry points. */
+/** Shared esbuild options for all entry points. */
 const shared = {
 	bundle: true,
 	format: "esm",
@@ -55,6 +55,14 @@ await Promise.all([
 		entryPoints: ["src/index.ts"],
 		outfile: "dist/index.js",
 	}),
+	// Hub daemon spawn target — re-exports @clinebot/core/hub/daemon-entry.
+	// The CLI spawns the daemon from dist/entry.js; without this entry point
+	// the published package crashes with MODULE_NOT_FOUND on startup (#495).
+	esbuild.build({
+		...shared,
+		entryPoints: ["src/entry.ts"],
+		outfile: "dist/entry.js",
+	}),
 ]);
 
-console.log("esbuild: bundled dist/cli.js and dist/index.js");
+console.log("esbuild: bundled dist/cli.js, dist/index.js, and dist/entry.js");
--- a/src/entry.ts
+++ b/src/entry.ts
@@ -1,0 +1,45 @@
+/**
+ * Hub daemon spawn target.
+ *
+ * The kanban CLI launches the hub daemon from `dist/entry.js` (resolved
+ * relative to the package root). The actual daemon implementation is shipped
+ * inside the embedded `@clinebot/core` SDK, whose `package.json` exports map
+ * exposes it as the public subpath `@clinebot/core/hub/daemon-entry`.
+ *
+ * This file is ESM (`kanban` package.json has `"type": "module"`). Do not use
+ * CommonJS `require()` / `module.exports` here — Node loads `.js` files in an
+ * ESM package as modules, so `require` is not defined.
+ */
+import { fileURLToPath } from "node:url";
+import { dirname, join } from "node:path";
+
+const __filename = fileURLToPath(import.meta.url);
+const __dirname = dirname(__filename);
+
+let mod: Record<string, unknown>;
+try {
+	// Primary: honor the @clinebot/core exports map. The deep path
+	// `@clinebot/core/dist/hub/daemon/entry.js` is blocked by the `exports`
+	// field, so we must use the public subpath.
+	mod = (await import("@clinebot/core/hub/daemon-entry")) as Record<string, unknown>;
+} catch {
+	// Fallback: resolve the nested file directly relative to this module's
+	// location, bypassing the exports map (for environments that don't honor it).
+	const candidate = join(
+		__dirname,
+		"..",
+		"node_modules",
+		"@clinebot",
+		"core",
+		"dist",
+		"hub",
+		"daemon",
+		"entry.js",
+	);
+	mod = (await import(candidate)) as Record<string, unknown>;
+}
+
+const defaultExport = (mod as { default?: unknown }).default ?? mod;
+
+export default defaultExport;
+export const __esModule = true;
--- a/src/entry.ts
+++ b/src/entry.ts
@@ -9,6 +9,15 @@
  * This file is ESM (`kanban` package.json has `"type": "module"`). Do not use
  * CommonJS `require()` / `module.exports` here — Node loads `.js` files in an
  * ESM package as modules, so `require` is not defined.
+ *
+ * Runtime bootstrap: the `@clinebot/core` daemon entry references a bare global
+ * `AI` (the Vercel AI SDK namespace) at module-evaluation time via an
+ * __exportStar-style helper (`C(L, AI)`). When this module runs as the
+ * hub-daemon main process (`node dist/entry.js`), that global is not defined,
+ * which surfaces as `ReferenceError: AI is not defined`. We therefore
+ * establish `globalThis.AI` (the installed `ai` package) here, before importing
+ * the core entry, so the daemon can load and start. This is a no-op when the
+ * global is already provided by a host launcher.
  */
 import { fileURLToPath } from "node:url";
 import { dirname, join } from "node:path";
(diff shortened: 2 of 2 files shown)


bradrushworth:
Pushed a fix to `src/entry.ts` on this branch — the shim was missing a required piece.

**Gap:** This PR correctly fixes the `MODULE_NOT_FOUND` half of #495 (adds `dist/entry.js` to the build + a third esbuild entry point). But a cold `node dist/entry.js` spawn **still crashes** with `ReferenceError: AI is not defined`. The embedded `@clinebot/core` daemon entry references a bare global `AI` (the Vercel `ai` SDK namespace) at module-evaluation time via an `__exportStar`-style helper (`C(L, AI)`). When the daemon runs as a standalone `node dist/entry.js` process, nothing defines that global, so it throws before the hub can start.

The previous "validation" (clean `hub-daemon.log`) was a false positive — it only proves the file resolves; a warm hub already bound to the port keeps the log clean regardless. On a fresh spawn with this shim as written, the daemon dies with `AI is not defined`, so the CLI would still hang.

**Fix:** `src/entry.ts` now establishes `globalThis.AI` (the installed `ai` package) **before** importing `@clinebot/core/hub/daemon-entry` (with a nested-path fallback), so the daemon can actually load and start. It's a no-op when a host launcher already provides the global.

Verified locally: with the bootstrap, `node dist/entry.js` loads the core entry and the hub binds/starts (only `EADDRINUSE` if another hub is already up — the healthy case). Without it, the same command throws `AI is not defined`.

This now fully resolves #495 (both the `MODULE_NOT_FOUND` and `AI is not defined` halves). Note it is independent of #581 (terminal scrollback UI freeze).

Issue #539, opened 2026-06-22
Terminal viewport jumps when scrolled back after agent output / restore /   resize


benfargher:
### What happened?

  When viewing a long Kanban terminal session, if I scroll up in the terminal history,
  the viewport sometimes jumps back to the top or much higher in the buffer when the
  agent emits output or the terminal reconnects/restores/resizes.

  Expected behavior:
  - If the terminal is pinned to the bottom, new output should keep it pinned to the new
  bottom.
  - If the user has scrolled back, Kanban should preserve that viewport across writes,
  restore replay, and fit/resize operations.

  Environment:
  - Cline Kanban v0.1.68
  - Web UI terminal pane

  Suspected cause:
  `web-ui/src/terminal/persistent-terminal-manager.ts` calls `terminal.reset()` /
  snapshot replay in `applyRestore()`, writes output asynchronously via
  `terminal.write(...)`, and calls `fitAddon.fit()` in `requestResize()`. These
  operations can mutate xterm’s viewport without preserving whether the user was bottom-
  pinned or intentionally scrolled back.

  Proposed fix:
  Capture the terminal viewport before writes, restore replay/reset, and fit/resize.
  Restore afterward:
  - if previously bottom-pinned, call `scrollToBottom()`
  - if previously scrolled back, call `scrollToLine(previousViewportY)`, clamped to the
  current buffer length

  I tested a local patch against v0.1.68 with:
  - `npm --prefix web-ui test -- terminal-viewport.test.ts`
  - `npm --prefix web-ui run typecheck`
  - `npm run build`

  I can provide the full patch if useful.

## Steps to reproduce

  1. Start Kanban and open a task that produces a long terminal session.
  2. Let the terminal produce enough output to create scrollback.
  3. Scroll up inside the terminal pane so you are no longer pinned to the bottom.
  4. Leave the task running so the agent emits more output.

  Optional triggers that also seem related:

  5. Switch away from the task and back again, or refresh/reconnect the browser tab.
  6. Resize the Kanban window or terminal pane.

  ## Actual behavior

  The terminal viewport sometimes jumps to the top, or at least much higher in the
  terminal history, even though I was intentionally scrolled back.

  

### Expected behavior

  If I am scrolled back, the terminal should preserve my current viewport when new
  output arrives, when the terminal restores, and when the pane resizes. If I am already
  pinned to the bottom, it should stay pinned to the new bottom.

### Kanban version

0.1.68

### Environment

[...]

Issue #542, opened 2026-06-25
Terminal output can freeze permanently when a viewer stops acknowledging PTY   output


benfargher:
### What happened?

A Codex task in Kanban continued running and modifying files, but the Kanban
  chat/terminal UI stopped updating. This happened multiple times in one day.

  For the affected task:

  - Codex process was still running.
  - Codex transcript file continued to update.
  - File diffs continued changing.
  - Kanban hook/state updates continued.
  - Kanban terminal lastOutputAt stopped moving.
  - A diagnostic WebSocket to /api/terminal/control received live state updates.
  - A diagnostic WebSocket to /api/terminal/io received no output chunks.

  This suggests the task/agent is healthy, but the terminal IO stream is paused
  or detached.

### Steps to reproduce

 1. Start a Codex task in Kanban that produces enough terminal output to
     trigger terminal WebSocket backpressure.

  2. Leave the browser/detail pane open long enough for the terminal renderer to
     stop acknowledging output, or otherwise wedge the viewer.

  3. Observe the agent continues running and hooks/state continue updating.
  4. Observe the Kanban terminal/chat output stops permanently for that task.
  5. Reconnecting a viewer may not recover raw IO, even though control state
     continues.

  

### Expected behavior

  A stale or wedged browser terminal viewer should not be able to keep the
  shared task PTY paused indefinitely. Kanban should either evict the non-acking
  viewer, force-resume after a timeout, or otherwise allow a fresh viewer to
  recover the stream.

### Kanban version

  0.1.68

### Environment

  Linux VPS, Node 22, Kanban served at 127.0.0.1:3484, Codex CLI 0.142.0.

bradrushworth:
Consolidated diagnosis of this terminal-freeze class (snapshot + 10k scrollback on the main thread during a task switch, plus the backpressure wedge) is now tracked in #581. Linking for visibility.

TyroneNel:
[...]

PR #583, opened 2026-08-06
fix(terminal): cap scrollback 10k→1k + viewport-only restore to stop task-switch freeze (#581)

### LINKED ISSUE #581: Web UI freezes ~60s (page unresponsive) when switching tasks during a long agent run (terminal snapshot + 10k scrollback on main thread)

bradrushworth:
## Summary

Switching tasks during a long agent run freezes the browser tab for ~60 seconds ("page unresponsive"). The freeze is dominated by the terminal restore/snapshot path: on every task switch the server serializes a **full headless xterm buffer** and ships it over the WebSocket, and the browser re-parses and re-renders the entire buffer in an xterm with a large scrollback.

Fixes #581 — implements the full **P0 remediation** (both "cap & slim the snapshot" **and** "viewport-only restore").

## Root cause

Two compounding problems on the restore path:

1. **Unbounded scrollback** — `TERMINAL_SCROLLBACK = 10_000` on **both** the server-side headless terminal mirror and the browser-side xterm:
   | Side | File | Before |
   |------|------|--------|
   | Server (headless mirror) | `src/terminal/terminal-state-mirror.ts` | `TERMINAL_SCROLLBACK = 10_000` |
   | Browser (xterm options) | `web-ui/src/terminal/terminal-options.ts` | `scrollback: 10_000` |

2. **Full-buffer serialize on every restore** — on task switch (`controlSocket` → `restore` frame), the server calls `getSnapshot()` → `serializeAddon.serialize()` (no args) which serializes the **entire** scrollback buffer into one string. The browser `applyRestore()` then does `terminal.reset()` + `terminal.write(fullSnapshot)`, blocking the main thread as xterm parses/renders every line.

Long/complex runs grow this buffer, so the freeze worsens over time — and the server-side `serialize()` competes with HTTP request handling on the same Node event loop. The worst case also drives the OOM trajectory in #273 (`JsonStringify → SerializeArrayLikeSlow`).

### Part 1 — cap scrollback 10,000 → 1,000 (both sides)
- **`src/terminal/terminal-state-mirror.ts`**: `TERMINAL_SCROLLBACK = 1_000`
- **`web-ui/src/terminal/terminal-options.ts`**: `scrollback: 1_000` (extracted to a `TERMINAL_SCROLLBACK` constant, kept in sync with the server)
- **`web-ui/src/terminal/terminal-options.test.ts`**: assertion `10_000` → `1_000`

1,000 lines is enough for interactive use (the visible viewport is ~24–50 rows; 1,000 ≈ 20–40 screenfuls of recent history) while capping the buffer 10× smaller. This alone cuts the snapshot/serialize/render cost by ~10×.

### Part 2 — viewport-only restore (`serialize({ scrollback: 0 })`)
- **`src/terminal/terminal-state-mirror.ts`** `getSnapshot()`: `serializeAddon.serialize()` → `serializeAddon.serialize({ scrollback: 0 })`

With `{ scrollback: 0 }`, the SerializeAddon emits **only the visible viewport** (≈ `rows × cols`) instead of the full 1,000-line buffer. Live agent output continues to stream incrementally over the WebSocket after restore, so no active-session history is lost — only the initial replay cost is removed.

This makes the restore payload **O(viewport)** instead of **O(scrollback)**, eliminating the residual restore-path freeze for worst-case long runs that fill the 1,000-line cap. It turns the ~6s residual (after Part 1) into sub-millisecond.

## Changes

- `src/terminal/terminal-state-mirror.ts`: `10_000` → `1_000`; `serialize()` → `serialize({ scrollback: 0 })`; explanatory comments
- `web-ui/src/terminal/terminal-options.ts`: `10_000` → `1_000` (extracted to `TERMINAL_SCROLLBACK` constant) + comment
- `web-ui/src/terminal/terminal-options.test.ts`: assertion `10_000` → `1_000`

## Why this is the complete P0 (not just Part 1)

Issue #581's remediation plan splits P0 into "cap & slim" and "viewport-only restore / incremental deltas". This PR does **both**, so the restore path is fixed end-to-end:

| Symptom | After Part 1 (cap only) | After Part 1 + Part 2 (this PR) |
|---|---|---|
| ~60s freeze on task switch | ~6s (10× smaller buffer) | sub-second / imperceptible (viewport only) |
| OOM trajectory (#273) | mitigated (10× smaller serialize) | mitigated + serialize is O(viewport) |
| Cost scaling with run length | still scales (fills 1k cap) | **constant** (viewport-sized) |

> **Note on lazy-loading older scrollback:** a fuller architecture would also lazy-load older scrollback on scroll (so users can still see >1,000 lines of history on demand). That's a larger protocol change and is left for a follow-up; this PR's `{ scrollback: 0 }` is the minimal, safe way to remove the restore-path freeze without protocol changes.

## What this does NOT do (other tiers from #581, left for follow-ups)

- ❌ **P1 — append-only `persistMessages()`** instead of full rewrite per `iteration_end` (separate code path, "worsens over time" cost)
- ❌ **P2 — cache `board.json`/`sessions.json`, async `readJsonFile`** (HTTP read path)
- ❌ **P3 — move `getSnapshot()` off the event loop (worker thread) / enforce backpressure (#542)**

## Verification

- `web-ui/src/terminal/terminal-options.test.ts` updated — `npm run web:test` should pass.
- The scrollback change is a pure constant; the `serialize({ scrollback: 0 })` change is a one-arg addition to an existing API call (`@xterm/addon-serialize` `ISerializeOptions.scrollback`). No type or API surface changes.
- Manual repro from the issue (click between tasks during a long agent run): the restore snapshot is now a few KB (viewport-sized) instead of up to ~1 MB, and `terminal.write(snapshot)` completes in well under a second.
- Validated independently on a live `kanban` v0.1.70 instance: applying both the 1k cap and `serialize({ scrollback: 0 })` to the installed `dist/cli.js` resolved the ~60s task-switch freeze, and the served web bundle confirmed the patched `scrollback: 1e3`.

## Related issues

- #581 — this PR (full P0 remediation)
- #273 — OOM crash on `getSnapshot().serialize()` — this PR makes serialize O(viewport) and 10× smaller, mitigating the OOM trajectory
- #542 — terminal output freezes permanently when a viewer stops acknowledging (backpressure wedge) — orthogonal, still open
- #539 / #541 — viewport jump on restore — orthogonal (unmerged PR #541 addresses that separately)

## Note on the previous PR body

An earlier version of this PR body referenced "2,000" in the prose. The committed code has always used **1,000** (the floor suggested in #581); this updated body matches the code. The code is the source of truth.

### DIFF
--- a/src/terminal/terminal-state-mirror.ts
+++ b/src/terminal/terminal-state-mirror.ts
@@ -4,7 +4,17 @@ import headlessTerminalModule from "@xterm/headless";
 const { SerializeAddon } = serializeAddonModule as typeof import("@xterm/addon-serialize");
 const { Terminal } = headlessTerminalModule as typeof import("@xterm/headless");
 
-const TERMINAL_SCROLLBACK = 10_000;
+/**
+ * Server-side headless terminal scrollback.
+ *
+ * Kept deliberately small because the entire buffer is serialized via
+ * `serializeAddon.serialize()` and shipped over the WebSocket on every
+ * viewer connect (task switch). A 10,000-line buffer caused ~60s main-thread
+ * freezes (#581) and RSS spikes toward OOM (#273) during long agent runs.
+ * 1,000 lines is enough to show recent output while keeping the restore
+ * snapshot ~10x smaller.
+ */
+const TERMINAL_SCROLLBACK = 1_000;
 
 export interface TerminalRestoreSnapshot {
 	snapshot: string;
--- a/src/terminal/terminal-state-mirror.ts
+++ b/src/terminal/terminal-state-mirror.ts
@@ -68,7 +68,14 @@ export class TerminalStateMirror {
 	async getSnapshot(): Promise<TerminalRestoreSnapshot> {
 		await this.operationQueue;
 		return {
-			snapshot: this.serializeAddon.serialize(),
+			// Serialize only the visible viewport (scrollback: 0) instead of the
+			// full buffer. The viewer immediately receives live output deltas
+			// after restore, so shipping the entire scrollback history on every
+			// task switch is unnecessary and was the dominant cause of the
+			// ~60s main-thread freeze (#581) and OOM trajectory (#273) during
+			// long agent runs. This keeps the restore payload to roughly
+			// rows*cols bytes regardless of run length.
+			snapshot: this.serializeAddon.serialize({ scrollback: 0 }),
 			cols: this.terminal.cols,
 			rows: this.terminal.rows,
 		};
--- a/web-ui/src/terminal/terminal-options.test.ts
+++ b/web-ui/src/terminal/terminal-options.test.ts
@@ -16,7 +16,7 @@ describe("createKanbanTerminalOptions", () => {
 		expect(options.cursorBlink).toBe(false);
 		expect(options.cursorInactiveStyle).toBe("outline");
 		expect(options.cursorStyle).toBe("block");
-		expect(options.scrollback).toBe(10_000);
+		expect(options.scrollback).toBe(1_000);
 		expect(options.macOptionIsMeta).toBe(true);
 		expect(options.windowOptions).toEqual({
 			getCellSizePixels: true,
--- a/web-ui/src/terminal/terminal-options.ts
+++ b/web-ui/src/terminal/terminal-options.ts
@@ -13,6 +13,18 @@ const TERMINAL_WORD_SEPARATOR = " ()[]{}',\"`";
 const TERMINAL_FONT_FAMILY =
 	"'Cascadia Code', 'Fira Code', 'JetBrains Mono', 'SF Mono', Menlo, Monaco, 'Courier New', monospace";
 
+/**
+ * Browser xterm scrollback.
+ *
+ * Must stay in sync with the server-side `TERMINAL_SCROLLBACK` in
+ * `src/terminal/terminal-state-mirror.ts`. On every task switch the browser
+ * receives a full `restore` snapshot and does `terminal.reset()` +
+ * `terminal.write(snapshot)`. A 10,000-line scrollback made this write block
+ * the main thread for ~60s during long agent runs (#581). 1,000 lines keeps
+ * recent output visible while capping the restore cost at ~10x smaller.
+ */
+const TERMINAL_SCROLLBACK = 1_000;
+
 export function createKanbanTerminalOptions({
 	cursorColor,
 	isMacPlatform,
@@ -38,7 +50,7 @@ export function createKanbanTerminalOptions({
 		rightClickSelectsWord: false,
 		scrollOnEraseInDisplay: true,
 		scrollOnUserInput: true,
-		scrollback: 10_000,
+		scrollback: TERMINAL_SCROLLBACK,
 		smoothScrollDuration: 0,
 		theme: {
 			background: terminalBackgroundColor,


bradrushworth:
**Correction to the PR body above:** The body text references "2,000" in several places, but the actual committed code uses **1,000** lines (`TERMINAL_SCROLLBACK = 1_000` on both the server and browser sides, and the test asserts `1_000`). The inline code comments in the diff are correct — they say 1,000. 1,000 is the value that was actually implemented and is the floor suggested in issue #581. Please disregard the "2,000" figures in the prose body; the code is the source of truth.

Summary of actual changes:
- `src/terminal/terminal-state-mirror.ts`: `10_000` → `1_000`
- `web-ui/src/terminal/terminal-options.ts`: `10_000` → `1_000` (extracted to a `TERMINAL_SCROLLBACK` constant)
- `web-ui/src/terminal/terminal-options.test.ts`: assertion `10_000` → `1_000`

TyroneNel:
Leaving this PR as-is. I did not want to push onto this branch.

The 1k cap is the right first cut. The later viewport-only `serialize({ scrollback: 0 })` commit drops everything above the current screen, so I did not take that path.

#596 keeps the 1k cap, caches the full serialize until output or resize, applies restore in 16 KiB writes, and skips reset/rewrite when the same PTY generation is already on the client.

Happy to close this in favor of that PR, or to leave both open if you want to compare.

PR #606, opened 2026-08-20
perf: use lightweight Codex hook entrypoint

Cascoopman:
## Summary

- route Codex lifecycle hooks through a dedicated lightweight entrypoint instead of booting the full Kanban CLI for every hook
- preserve the existing main-CLI fallback for opaque executable launches
- build and desktop-stage the new `dist/codex-hook.js` artifact
- add a reusable process monitor for runtime, hook, Git, agent, and terminal resource usage

## Root cause

Every Codex `PreToolUse` and `PostToolUse` hook launched the full Kanban CLI through `tsx`. That transitively loaded runtime telemetry, Sentry, and server dependencies for each tool call, causing repeated high-memory Node processes and contention during normal agent activity.

Observed before the change:

- individual hooks: 409-445 MiB RSS and 0.78-1.99s
- contended full-CLI load: 418 MiB RSS and 5.38s
- 25s trace: 20 hook processes, 125% average CPU, 540% peak CPU, 899 MiB combined peak RSS
- trivial tool commands: 2.5-6.3s end-to-end

With the dedicated entrypoint:

- source hook: 0.39-0.60s and about 104 MiB RSS
- bundled hook: 0.11-0.12s and about 67 MiB RSS
- roughly 84% lower peak RSS for the production hook path

## Validation

- backend and web TypeScript checks pass
- lint passes
- fast suite passes: 54 files / 554 tests
- focused hook tests pass: 34 tests
- production build passes and emits the 839 KiB hook bundle alongside the 16 MiB main CLI
- desktop staging includes and validates `codex-hook.js`
- local production smoke test on port 3499: UI HTTP 200, passcode endpoint HTTP 200, title `Kanban`, performance monitor collected samples, and shutdown completed cleanly

## Follow-up work

The investigation also found two independent performance areas that are intentionally not included here: very large diff rendering can create hundreds of thousands of DOM nodes, and workspace metadata polling can spawn a high volume of Git processes. Those should be addressed separately to keep this change focused.

### DIFF
--- a/package.json
+++ b/package.json
@@ -60,6 +60,7 @@
     "typecheck": "tsc -p tsconfig.json --noEmit",
     "lint": "npx @biomejs/biome lint src test web-ui/src web-ui/tests vitest.config.ts web-ui/vite.config.ts web-ui/vitest.config.ts web-ui/playwright.config.ts",
     "format": "npx @biomejs/biome check --write .",
+    "perf:monitor": "node scripts/monitor-performance.mjs",
     "check": "npx @biomejs/biome check . && npm run typecheck && npm run test",
     "test": "vitest run",
     "test:fast": "vitest run test/runtime test/utilities",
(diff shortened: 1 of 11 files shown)


Cascoopman:
Closing because this PR was opened against the wrong repository. The change is intended for Cascoopman/kanban.

PR #612, opened 2026-08-21
fix(runtime-state-hub): heartbeat to drop dead WebSocket clients (#285)

### LINKED ISSUE #285: WebSocket server lacks ping/pong heartbeat → dead clients accumulate
## Problem

The runtime-state WebSocket server (`src/server/runtime-state-hub.ts` in pre-0.1.59 layout, now bundled into `dist/cli.js`) does not ping connected clients. When a client's TCP connection silently dies — laptop sleep, network blip, browser tab killed before clean unload — the server never notices. The stale client stays in the internal \`runtimeStateClients\` Set forever and continues to receive broadcasts that go nowhere.

Over time this causes:

- Memory growth as dead clients accumulate across the day.
- Wasted broadcast work in `broadcastRuntimeProjectsUpdated`, `broadcastClineMcpAuthStatusesUpdated`, etc.
- Potential event-loop pressure on servers with many short-lived sessions.

## Proposed fix

Add a standard WebSocket ping/pong heartbeat. Sketch (bundler-friendly, no new deps):

```ts
const RUNTIME_STATE_WS_HEARTBEAT_MS = 20_000;

const markAlive = (client: WebSocket & { isAlive?: boolean }) => {
  client.isAlive = true;
};

const heartbeatTimer = setInterval(() => {
  for (const client of runtimeStateClients) {
    if (client.readyState !== WebSocket.OPEN) continue;
    if (client.isAlive === false) {
      try { client.terminate(); } catch { /* ignore */ }
      continue;
    }
    client.isAlive = false;
    try { client.ping(); } catch {
      try { client.terminate(); } catch { /* ignore */ }
    }
  }
}, RUNTIME_STATE_WS_HEARTBEAT_MS);
heartbeatTimer.unref?.();

runtimeStateWebSocketServer.on("connection", async (client, context) => {
  markAlive(client);
  client.on("pong", () => markAlive(client));
  client.on("close", () => {
    client.removeAllListeners?.("pong");
    cleanupRuntimeStateClient(client);
  });
  // ... existing handler
});

// In shutdown cleanup:
clearInterval(heartbeatTimer);
```

## Why I care

I've been running a local monkey-patch that injects exactly this logic into `dist/cli.js` on every `npm install -g kanban`. It works but breaks on every bundler refactor (most recently 0.1.49 → 0.1.50 shifted anchors, 0.1.58 → 0.1.59 moved the whole file into the bundled cli.js). A native implementation would let me delete ~100 lines of patching infrastructure.

Happy to open a PR if there's interest.

## Environment

- `kanban@0.1.59`
- Long-running single-user board via systemd on a Hetzner VPS

pacocartones:
## Problem

Fixes #285. The runtime-state WebSocket hub (`src/server/runtime-state-hub.ts`) never pings its clients. When a client's TCP connection dies silently (laptop sleep, network blip, a browser tab killed before a clean unload), the server never notices: the stale socket stays in `runtimeStateClients` forever and keeps receiving writes.

This is distinct from the terminal-PTY backpressure work in #607/#610/#611 (`src/terminal/ws-server.ts`); this PR only touches the runtime-state hub.

## Fix

Add a ping/pong heartbeat sweep in `createRuntimeStateHub`:

- On each connection, mark the socket alive and refresh liveness on every `pong`.
- Every interval (default **30s**), terminate and clean up sockets that did not answer the previous ping, then ping the rest.
- The interval is configurable via an optional `heartbeatIntervalMs` dep (used by the test), and the timer is `unref()`-ed so it never keeps the process alive on its own.
- `clearInterval` on hub `close()`.

## Tests

Adds `test/server/runtime-state-hub-heartbeat.test.ts` with two cases against a real ephemeral server:

- a client that never answers the ping is terminated;
- a client that answers pings stays connected.

Verified red→green: without the fix the dead-client case fails; with it both pass. No regressions in `test:fast`. `tsc --noEmit` clean.

## Scope

Bug fix / stability only. No protocol or message-shape changes; standard WebSocket control frames.

### DIFF
--- a/src/server/runtime-state-hub.ts
+++ b/src/server/runtime-state-hub.ts
@@ -36,6 +36,9 @@ export interface CreateRuntimeStateHubDependencies {
 		WorkspaceRegistry,
 		"resolveWorkspaceForStream" | "buildProjectsPayload" | "buildWorkspaceStateSnapshot"
 	>;
+	// #285: optional heartbeat interval override (milliseconds). Defaults to 30s.
+	// Exposed so tests can drive the ping/pong sweep on a short real interval.
+	heartbeatIntervalMs?: number;
 }
 
 export interface RuntimeStateHub {
@@ -72,6 +75,34 @@ export function createRuntimeStateHub(deps: CreateRuntimeStateHubDependencies): 
 	const runtimeStateWorkspaceIdByClient = new Map<WebSocket, string>();
 	let clineSessionContextVersion = 0;
 	const runtimeStateWebSocketServer = new WebSocketServer({ noServer: true });
+	// #285: detect dead clients (laptop sleep, network blip, tab killed before clean
+	// unload) with a WebSocket ping/pong heartbeat, so stale sockets do not accumulate
+	// in runtimeStateClients forever. Unresponsive sockets are terminated and cleaned up.
+	const RUNTIME_STATE_HEARTBEAT_INTERVAL_MS = deps.heartbeatIntervalMs ?? 30_000;
+	const runtimeStateClientLiveness = new WeakMap<WebSocket, boolean>();
+	const runtimeStateHeartbeatTimer = setInterval(() => {
+		for (const client of runtimeStateClients) {
+			if (runtimeStateClientLiveness.get(client) === false) {
+				cleanupRuntimeStateClient(client);
+				try {
+					client.terminate();
+				} catch {
+					// Ignore termination errors for an already-dead socket.
+				}
+				continue;
+			}
+			runtimeStateClientLiveness.set(client, false);
+			try {
+				client.ping();
+			} catch {
+				// Ignore ping errors; the next sweep terminates the socket.
+			}
+		}
+	}, RUNTIME_STATE_HEARTBEAT_INTERVAL_MS);
+	// Do not keep the process alive solely for the heartbeat sweep.
+	if (typeof runtimeStateHeartbeatTimer.unref === "function") {
+		runtimeStateHeartbeatTimer.unref();
+	}
 	const workspaceMetadataMonitor = createWorkspaceMetadataMonitor({
 		onMetadataUpdated: (workspaceId, workspaceMetadata) => {
 			const clients = runtimeStateClientsByWorkspaceId.get(workspaceId);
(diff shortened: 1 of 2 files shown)


TyroneNel:
Nice fix — this closes a real gap. The runtime-state hub genuinely had no liveness check, and dead sockets sitting in `runtimeStateClients` and still taking writes is exactly the leak #285 describes. Thank you for picking it up.

Two things from testing your branch.

## 1. `tsc` fails on the new test

I checked out your branch on its own, against a clean `main`, and ran the full check. `tsc` reports:

```
test/server/runtime-state-hub-heartbeat.test.ts(33,41): error TS2345:
Argument of type 'undefined' is not assignable to parameter of type
'{ requestedWorkspaceId: string | null; }'
```

Line 33 is this call:

```ts
hub.handleUpgrade(req, socket, head, undefined);
```

The context parameter is required, and it cannot be `undefined`. Passing an object should fix it:

```ts
hub.handleUpgrade(req, socket, head, { requestedWorkspaceId: null });
```

The tests still pass, because Vitest does not type-check. But `npm run check` runs `tsc`, so CI will go red. I think you may have run a narrower check than the full gate.

Everything else looks good. I merged your branch into my fork on top of the terminal work in #607 and #611. It merged cleanly, no conflicts, and all 128 tests in `test/server/` and `test/runtime/terminal/` passed together.

## 2. Your fix and #607 are complementary, not overlapping

You already noted this in the description, and I want to confirm it and add one detail that took me a while to find.

A ping/pong heartbeat catches a socket whose **TCP connection is dead**. That is the laptop-sleep and network-blip case, and your fix handles it.

It does not catch a **backgrounded browser tab**. Browsers answer ping frames in the network stack, without running any JavaScript. So a suspended tab keeps sending pongs and looks perfectly healthy to a heartbeat, even though its renderer is frozen and cannot do anything.

That is why #607 needed a second signal. On the terminal socket the frozen tab kept ponging while it stopped acknowledging output, so it held the PTY paused and blocked the agent process on its next write to stdout. The fix there watches acknowledgement *progress* rather than socket liveness.

This does not affect your PR. The runtime-state hub only pushes board state, so a frozen tab there wastes bandwidth instead of blocking anything. But it is worth knowing if anyone later suggests one heartbeat could cover both hubs. It cannot — the two sockets fail in different ways.

Happy to review a follow-up if you take on more of this area.

pacocartones:
Thanks for catching that — you were right that the focused Vitest run missed the required upgrade context type. Fixed in 8582017 by passing `{ requestedWorkspaceId: null }`. I reran the full TypeScript check and the focused heartbeat test: `npm run typecheck` passes and the test is 2/2 green. Also agreed on the distinction from #607: this heartbeat detects dead transports, while acknowledgement progress is the signal for a live socket whose renderer is suspended.

PR #628, opened 2026-08-28
feat: add project-scoped task memory control

d0uub:
## Concept

This introduces layered memory control so agents can reuse prior work without injecting every task transcript into every future session:

1. **Card summaries** are concise, task-owned handoffs. Cline drafts one from the final response, users can edit it, and failed attempts retain what was tried and why work stopped.
2. **Project memory** is curated, durable guidance for architecture facts, commands, conventions, and known failure prevention. Users can edit it directly or promote a card summary.
3. **Task memory index** is generated project context. It contains bounded previews for summarized active, completed, failed, and stopped tasks, plus one hashed Markdown detail file per task for on-demand drill-down.
4. **Dependency context** injects summaries from completed prerequisites into linked tasks, avoiding broad unrelated context.

## Lifecycle

- state saves refresh memory for summarized cards that remain on the board
- clearing Done or deleting a batch archives all target tasks before board removal
- tasks without an outcome summary receive an explicit fallback containing the original request rather than disappearing silently
- failed and interrupted work is labeled separately from successful completion
- every new native Cline session receives curated project memory plus the compact task index; details are read only when relevant

## Controls and safety

- project memory is workspace-scoped and limited to 10,000 characters
- card summaries are limited to 2,000 characters
- generated task indexes are bounded to 8,000 characters
- project-memory promotion uses a locked atomic read-modify-write
- project-memory APIs require a resolved workspace scope
- task detail paths are derived from SHA-256 task-ID hashes and never trusted from the persisted manifest
- generated task summaries are explicitly marked as historical reference, not instructions

## Validation

- 99 focused tests passed across card summaries, project memory, task memory, batch deletion, dependency summaries, and Cline prompt injection
- Biome passed for all 25 PR files
- editor diagnostics are clean
- `npm run build` passed

## Scope

This PR is independent of the custom-provider fix in #627 and contains no provider, development-script, KB-extension, or unrelated formatting changes.

### DIFF
--- a/docs/architecture.md
+++ b/docs/architecture.md
@@ -409,3 +409,32 @@ A new engineer opening this repo will probably notice a few things quickly:
 - the architecture now favors clean ownership over compatibility glue because this area did not have legacy users to preserve
 
 If you approach the code with those assumptions, the rest of the system starts to make sense much faster.
+
+## Project Memory
+
+Project memory is a per-project, user-authored context file (`project-memory.md`) injected into every new Cline task session.
+
+| Property | Value |
+| --- | --- |
+| Location | `~/.cline/kanban/workspaces/<workspaceId>/project-memory.md` |
+| Scope | per-project (workspace-scoped) |
+| Editing | manual, user-approved via Project Memory Editor dialog |
+| Size limit | 10,000 characters |
+| Injection | automatic into new Cline sessions |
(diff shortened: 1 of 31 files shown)


d0uub:
Withdrawing this PR because the original memory implementation was incomplete. The completed work is preserved on the branch in commit 0e84ec1 for follow-up; this PR should not be merged in its current form.

